PR #3, opened 2015-01-15
'fs' => 'afero'

wayneeseguin:
Changed remaining 'fs' to 'afero' in README.


spf13:
thanks @wayneeseguin . Sorry this sat for so long. Merged as 139e50e

Issue #4, opened 2015-03-22
InMemFile.Read has inconsistent behaviour at EOF with (*os.File).Read


jaqx0r:
http://golang.org/pkg/os/#File.Read says
Read reads up to len(b) bytes from the File. It returns the number of bytes read and an error, if any. EOF is signaled by a zero count with err set to io.EOF.

I don't see this behaviour when using InMemFile to fake a filesystem; InMemFile returns the EOF in the last read of nonzero number of bytes before EOF.

Here's a short program, `afero.go`

    package main
    
    import (
    	"fmt"
    	"io/ioutil"
    	"os"
    	"path"
    
    	"github.com/spf13/afero"
    )

    type file interface {
    	Read([]byte) (int, error)
    }

    func read(f file) {
    	b := make([]byte, 4096)
    	n, err := f.Read(b)
    	fmt.Printf("n:%v, err:%v\n", n, err)
    	n, err = f.Read(b)
    	fmt.Printf("n:%v, err:%v\n", n, err)
    }

    func main() {
    	f := afero.MemFileCreate("foo")
    	f.WriteString("a\nb\nc\nd\n")
    	f.Seek(0, 0)
    	read(f)

    	workdir, err := ioutil.TempDir("", "eof_test")
    	if err != nil {
    		fmt.Printf("couldn't open tempdir")
    	}
        defer os.RemoveAll(workdir)
    	g, err := os.Create(path.Join(workdir, "foo"))
    	g.WriteString("a\nb\nc\nd\n")
    	g.Seek(0, 0)
    
    	read(g)
    }

Output of the program is:

    go run afero.go
    n:8, err:EOF
    n:0, err:EOF
    n:8, err:<nil>
    n:0, err:EOF

You can see in the second pair of lines the normal EOF behaviour is to return the number of bytes, and nil; then in the second read, no bytes read and the EOF error.

InMemFile returns EOF in both calls.

jaqx0r:
Fixed in jaqx0r/afero master.

PR #5, opened 2015-03-22
Fix InMemoryFile.Read to not return nonzero bytes and EOF in the same call.

jaqx0r:
Closes: #4


jaqx0r:
Ping, @spf13

spf13:
Merged as 59eceea. Thanks @jaqx0r

PR #10, opened 2015-06-11
[MemMap] Fix a data race

noxiouz:
There's an unprotected concurrent read access to the map. It leads to a datarace:
https://gist.github.com/noxiouz/afa65f5e1f12ee904268
 in test 
https://gist.github.com/noxiouz/33688e01d9d295d8d53d


spf13:
Merged as 5849bf9.

PR #12, opened 2015-08-08
Fix MemMapFs.Remove() to really delete the file.

alext:
It was attempting to delete a file with a hardcoded path of "name" as
opposed to the path in the `name` variable.

Fixing this exposed a deadlock because the function was attempting to
acquire an exclusive lock when it already had a read lock.

This PR also makes the function return an error for a non-existent file.


spf13:
Merged as 6ced24d

PR #16, opened 2015-10-17
fix Readdir() and Readdirnames()

### LINKED ISSUE #7: Bug in Mem file implementation for ReadDir/ReadDirNames?
The following code
```
	fs := &afero.MemMapFs{}
	if err := fs.MkdirAll("/a/b", 0777); err != nil {
		log.Fatalf("unable to create dir:%s", err)
	}
	log.Println(fs)

	infos, err := afero.ReadDir("/", fs)
	if err != nil {
		log.Fatalf("Error with Readdir:%s", err)
	}
	log.Println(infos)
```
prints a nil slice for contents of `/` rathern than a fileinfo for `/a`. The MkdirAll seems to work
```
	2015/03/23 09:09:13 &{map[/:0xc20805a120 /a/b:0xc20805a060 /a:0xc20805a0c0] 0xc20801e400}
	2015/03/23 09:09:13 []
```

mbertschler:
fixes #7 

I first had to fix registerWithParent(), unRegisterWithParent() and findParent() to correctly use the MemDir of the parent directory. Then I added versions of Open() and Mkdir() that don't acquire a lock, that they can be used atomically during Create() and Mkdir(). 
Finally I changed the output of the InMemoryFileInfo.Name() and Readdirnames() functions to only return the last element of the path, as does the os package.
I also added some tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,8 +100,10 @@ the ability to drop in other filesystems as desired.
 Then throughout your functions and methods use the methods familiar
 already from the OS package.
 
-Methods Available:
+File System Methods Available:
 
+       Chmod(name string, mode os.FileMode) : error
+       Chtimes(name string, atime time.Time, mtime time.Time) : error
        Create(name string) : File, error
        Mkdir(name string, perm os.FileMode) : error
        MkdirAll(path string, perm os.FileMode) : error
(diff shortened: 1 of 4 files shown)


boutros:
Would be great to have this merged

boutros:
I tested your implementation a bit, seems I can't get Readdir to return io.EOF when all directory items are read?

mbertschler:
Oops, I totally forgot about the Readdir(n) n>0 usecase. Only n<0 works like it should. I will try to fix this today.

mbertschler:
@boutros could you check if it works right for you now? I think I fixed all the problems with the latest commit. Maybe we should mention Readdir() and Readdirnames() in the Readme too before this is merged.

boutros:
Yes, it works as expected now @mbertschler  :)

mbertschler:
Thanks for the confirmation :)
@spf13 is there anything blocking this getting merged?

PR #13, opened 2015-09-08
Support O_CREATE, O_APPEND and O_TRUNC flags in MemMapFs.OpenFile

alext:
This adds support for the `O_CREATE`, `O_APPEND` and `O_TRUNC` flags to OpenFile. This improves parity with `os.OpenFile`, and makes it easier to unit test code that uses these flags (eg logging code.)

PR #17, opened 2015-11-05
Add Walk function, File.Sync, update README

### LINKED ISSUE #6: Any specific reason a Walk function is not implemented within the package?
I can see that Fs interface doesn't specify LStat which is used in the implementation of Walk in the standard library so any Walk implementation on afero would not be 100% compatible with the standard library

However, Walk is a very common requirement for filesystems and so I wonder if there's a specific reason afero chooses not to provide one. If no, would it be ok for me to add an implementation & file a pull request?

### LINKED ISSUE #14: File interface is missing Sync() method

### LINKED ISSUE #15: MemMapFs.OpenFile should handle O_CREATE flag
OpenFile just calls Open to virtually open a file read-only with no parsing of the flags.
At a minimum it should at least see if O_CREATE is one of the flags passed and create the file instead.

mbertschler:
Add a Walk function similar to filepath.Walk and write a test for it.
Fixes #6 

Add Sync function to the File interface and implement it for InMemoryFile.
Fixes #14 

Readme:
Add a new version 0.9.0 to Release Notes to highlight recent improvements in afero.
#13 by @alext also fixed #15 
Add mbertschler to Contributors.
Rename MemFile to its actual name InMemoryFile.
Add GoDoc badge and use svg travis badge.
Sort File interface alphabetically and add Sync().

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,12 +56,12 @@ mocking and to speed up unnecessary disk io when persistence isn’t
 necessary. It is fully concurrent and will work within go routines
 safely.
 
-### MemFile
+#### InMemoryFile
 
 As part of MemMapFs, Afero also provides an atomic, fully concurrent memory
 backed file implementation. This can be used in other memory backed file
 systems with ease. Plans are to add a radix tree memory stored file
-system using MemFile. 
+system using InMemoryFile. 
 
 ## Usage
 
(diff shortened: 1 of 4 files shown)


spf13:
merged as 1407a13

Issue #18, opened 2015-11-06
Regression in Hugo


bep:
Hugo builds started to fail recently due to an issue with Afero:

```
links
--- PASS: TestSiteInfoPermalinks (0.00s)
=== RUN TestSitemapOutput
--- FAIL: TestSitemapOutput (0.01s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x20 pc=0x660933]
 
goroutine 3538 [running]:
testing.func·006()
	c:/go/src/testing/testing.go:441 +0x188
github.com/spf13/afero.(*MemMapFs).registerWithParent(0xc082582fa0, 0x1679a0, 0xc082103c70)
	C:/GOPATH/src/github.com/spf13/afero/memmap.go:143 +0x463
github.com/spf13/afero.(*MemMapFs).Create(0xc082582fa0, 0xc0828368ba, 0x1, 0x0, 0x0, 0x0, 0x0)
	C:/GOPATH/src/github.com/spf13/afero/memmap.go:93 +0x19a
github.com/spf13/hugo/helpers.WriteToDisk(0xc0828368ba, 0x1, 0x167b60, 0xc082367500, 0x162ca0, 0xc082582fa0, 0x0, 0x0)
	C:/GOPATH/src/github.com/spf13/hugo/helpers/path.go:555 +0x198
github.com/spf13/hugo/target.(*Filesystem).Publish(0xc08232a5c0, 0x0, 0x0, 0x167b60, 0xc082367500, 0x0, 0x0)
	C:/GOPATH/src/github.com/spf13/hugo/target/file.go:39 +0xce
github.com/spf13/hugo/hugolib.(*Site).WriteDestFile(0xc082864d20, 0x0, 0x0, 0x167b60, 0xc082367500, 0x0, 0x0)
	github.com/spf13/hugo/hugolib/_test/_obj_test/site.go:1946 +0x1e8
github.com/spf13/hugo/hugolib.(*Site).renderAndWriteXML(0xc082864d20, 0xb4def0, 0xc, 0x0, 0x0, 0xace0c0, 0xc0827c8c60, 0xc0827cb290, 0x3, 0x3, ...)
	github.com/spf13/hugo/hugolib/_test/_obj_test/site.go:1772 +0x40a
[...]

Issue #20, opened 2015-11-07
Afero can't handle relative paths properly


mbertschler:
This code:
``` go
package main

import (
	"fmt"
	"os"
	"path/filepath"

	"github.com/spf13/afero"
)

func main() {
	fs := new(afero.MemMapFs)

	f, _ := fs.Create(filepath.Join(".", "file"))
	f.WriteString("TestFile content")
	f.Close()

	wd, _ := os.Getwd()

	_, err := fs.Open(filepath.Join(wd, "file"))
	if err != nil {
		fmt.Println(err)
	}

	fmt.Println("wd:", wd)

	fs.List()
}
```

Produces this output:
```
$ absrel
file does not exist
wd: /tmp
file 16
. 42
```

It doesn't make sense to save relative paths into the MemMapFs, they all should be absolute.

I am working on a fix for that problem, but I am not sure if this can work together with the working directory from the os package. The problem is that if I use `os.Chdir("/tmp/afero")`, and then create a file "./file" on OS X it is actually put into "/private/tmp/afero". So if I use the dir string that I used in the Chdir function, the file is also not found. I also can't always use `filepath.EvalSymlinks` because files that exist in a MemMapFs won't exist in the real file system.

moorereason:
The current behavior makes complete sense to me.  The `MemMapFs` starts from scratch at the root of a new, clean filesystem.  The "working directory" is always `/`.  Creating `./file` should create `/file`.  Relative paths appear to work just fine.

`MemMapFs` does not pay attention to the OS working directory (technically, neither does `OsFs` except for the fact that the `os` stdlib package takes care of it for you).  I think what you're really wanting is to add `Chdir` and `Getwd` to the `afero.Fs` interface.

Also, I'm confused by your last paragraph.  Are you expecting the `MemMapFs` to act as an overlay on the current OS filesystem?  If so, I think you're asking for #39.

mbertschler:
This bug was filed when afero tried to open the "." dir. The main problem here has actually been solved.
I thought of a feature where folders in `MemMapFs` would mirror os folders, so that for example /tmp on OSX would become /private/tmp. The thing is I am not sure if that makes sense anymore of if that would be something that actually benefits a real world program. 

As the main problem here is solved I am closing this.

PR #21, opened 2015-11-07
Quick fix for the bug where the "." file is not a directory

mbertschler:
memDir is only nil when it was not made with Mkdir or lockfreeMkdir. In this case the parent is also not a real directory. This currently only happens for the file "."

This is a quick hack to make the library usable again and fixes https://github.com/spf13/afero/issues/18

This is only a quick fix to make the library usable again. I am working on a proper solution that can handle absolute and relative paths interchangeable, which is not possible right now. The real problem is described here: https://github.com/spf13/afero/issues/20

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -130,16 +130,26 @@ func (m *MemMapFs) registerWithParent(f File) {
 		pdir := filepath.Dir(path.Clean(f.Name()))
 		err := m.lockfreeMkdir(pdir, 0777)
 		if err != nil {
-			log.Println("Mkdir error:", err)
+			//log.Println("Mkdir error:", err)
 			return
 		}
 		parent, err = m.lockfreeOpen(pdir)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


bep:
Your code looks great, but for the future, it is better if you could create separate branches (cheap in the Git world) -- and stop creating pull requests from your master. This PR will eventually look OK when merged, but it kind of looks a little bit messy as is.

bep:
Merged in 0a58b82

mbertschler:
Sorry, I was a bit lazy and still don't fully get how pull requests work in detail. For example why six commits show up here when there was actually only one that changes something (the other ones were already merged).
In the future I will use branches.
@spf13 Thanks for the push access :)

spf13:
I'm happy to teach you how best to manage things and use git to ensure a
clean change set. I'll email some directions and we can chat online if you
have any questions. One key is to rebase instead of merge.

PR #23, opened 2015-11-26
atomic requires struct field to be 64-bit aligned

tpng:
fix https://github.com/spf13/hugo/issues/1621

### DIFF
--- a/memfile.go
+++ b/memfile.go
@@ -34,16 +34,18 @@ type MemDir interface {
 }
 
 type InMemoryFile struct {
-	sync.Mutex
+	// atomic requires 64-bit alignment for struct field access
 	at           int64
-	name         string
-	data         []byte
-	memDir       MemDir
-	dir          bool
-	closed       bool
-	mode         os.FileMode
-	modtime      time.Time
 	readDirCount int64
+
+	sync.Mutex
+	name    string
+	data    []byte
+	memDir  MemDir
+	dir     bool
+	closed  bool
+	mode    os.FileMode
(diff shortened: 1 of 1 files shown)


mbertschler:
Interesting issue, thank you for the fix.

moorereason:
@tpng, I'm impressed.  How in the world did you come up with that?  What lead you to this solution?

tpng:
@moorereason I first encountered this issue with the new v0.15 hugo using the server command and found that only the 386 version has this issue.

Then I went adding log.Print in the source code to find where hugo hang and it ends up in [(*InMemoryFile).Open()](https://github.com/spf13/afero/blob/master/memfile.go#L57).

After that I try to create a [repro](https://play.golang.org/p/HvLjorfv5K) to see how atomic failed in this case, which crash with a nil pointer error on the call to `atomic.StoreInt64(&f.readDirCount, 0)`.

Finally a google search of `golang atomic 386 amd64` ended up with a [result that stated that struct field has to be 64-bit aligned](https://groups.google.com/forum/#!msg/golang-dev/LQnKdFNFWbo/MAlWkTFYV40J) and thus the fix.

moorereason:
That's the first time I've seen a ["Bugs" section](https://golang.org/pkg/sync/atomic/) in the Go stdlib docs.  The docs also say this affects non-Linux ARM platforms.

I'm not very familiar with AppVeyor, but it seems like afero could use AppVeyor to automate testing on the Windows x86 platform to catch this sneaky bug.

PR #25, opened 2015-12-02
Use filepath.Abs to interpret relative file paths

pwplus:
This should fix the problem mentioned in https://github.com/spf13/afero/issues/20. It does so by using filepath.Abs to interpret the relative file path (if there is one). It then registers it as the absolute path, rather than the relative path. This passes the test case written in the issue above.

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -88,6 +88,7 @@ func (MemMapFs) Name() string { return "MemMapFS" }
 
 func (m *MemMapFs) Create(name string) (File, error) {
 	m.lock()
+	name, _ = filepath.Abs(name)
 	file := MemFileCreate(name)
 	m.getData()[name] = file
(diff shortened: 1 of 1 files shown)


pwplus:
Manager approved and I have signed the CLA.

spf13:
@mbertschler can you look at this since it replaces code you wrote?

mbertschler:
@pwplus thanks for this PR but I think this is only a partial fix to the relative files problem. Such a fix also needs to be in other functions that create files like Mkdir.

@spf13 I am also not so sure if we should rely on the os working directory functions. If you would call os.Chdir on a folder that exists in a memory fs but not in the real fs it would not work.

I am still working on a proper fix for this relative problem on a seperate branch and hope to have something ready soon.

PR #32, opened 2015-12-07
Adding Afero Utils

spf13:
@mbertschler This is a big addition with two breaking changes (in https://github.com/spf13/afero/commit/2affd36ebaaa0b17bfef4101ccf89d3caea2a776)  to the existing utilities. 

Could you please review this. I recognize it won't merge cleanly due to some of your latest additions. I'm just submitting a PR to get you to review it and I'll merge and push it once you have. 

Thanks!

### DIFF
--- a/afero.go
+++ b/afero.go
@@ -30,7 +30,7 @@ import (
 )
 
 type Afero struct {
-	fs Fs
+	Fs
 }
 
 // File represents a file in the filesystem.
--- a/afero.go
+++ b/afero.go
@@ -26,10 +26,12 @@ import (
 	"errors"
 	"io"
 	"os"
-	"path/filepath"
-	"sort"
 	"time"
 )
+
+type Afero struct {
+	fs Fs
+}
 
 // File represents a file in the filesystem.
 type File interface {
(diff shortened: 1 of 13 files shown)


mbertschler:
@spf13 When I implemented the `Walk` function I even thought a bit about that but then just went with adopting Fs at the end like it was already with the other functions. I fully agree that having Fs as the first argument makes most sense.

I also think the additions of these utilities make sense for afero.

There is one thing I don't fully understand though. Why did you create this type?
```go
type Afero struct {
        fs Fs
}
```
Did you mean this?
```go
type Afero struct {
        Fs
}
```
That way a user could call all utility functions as methods of this and also call the Fs function directly? Am I missing something here? The current struct can't be used by users because there is no way to create an Afero object.

While looking at afero in godoc I thought about the overall exported API, but that is a seperate issue: #33

spf13:
Oh. I think you are right. I'll redo that and repush.

spf13:
I've manually merged this in. Thanks for the feedback.

mbertschler:
Thank you for this tools. I will definitely update my app as I prefer the use of for example `appFs.Walk()` to `Walk(appFs,...)`

PR #40, opened 2015-12-13
implement io.WriterTo on memFile

chonthu:


chonthu:
Sorry, just noticed your io helper functions for ioutil, closing.

chonthu:
Actually no I cant seem to do a io.Copy on a memfile without this.

mem/file.go
```diff
@@ -29,6 +29,8 @@ import "time"
 const FilePathSeparator = string(filepath.Separator)
 
 type File struct {
+	io.Writer
```
davecheney: This adds a public field, File.Writer to the type, which is unpopulated.

mem/file.go
```diff
@@ -218,6 +220,12 @@ func (f *File) WriteString(s string) (ret int, err error) {
 	return f.Write([]byte(s))
 }
 
+func (f *File) WriteTo(w io.Writer) (n int64, err error) {
```
davecheney: This is not the signature for io.Writer, this is signature for io.WriterTo

davecheney:
not lgtm sorry. The description doesn't match the change you've made in the code. I think you should raise an issue to discuss the problem you have before starting to code a solution.

Also, please include a test.

chonthu:
Will close and discuss in issues first.

PR #46, opened 2015-12-21
add the regexp filter

vetinari:
filter reworked:
- any FilterFs is now responsible to call the underlying Fs if it does not intercept and block the call
- add regexp filter to show only files matching the regexp, directories are passed as is


spf13:
merged as e7cb826

Issue #47, opened 2015-12-22
Initial sftp backend draft


xor-gate:
Hi guys,

I'm fairly new with golang and trying to get my hands dirty and have got a initial working draft on a SFTP backend.

**What works**
* Connecting/disconnecting (with $HOME/.ssh/id_rsa)
* Creating and writing a file
* Most trivial functions

**What works not**
* Passing advanced SSH/sftp config

Here is a example:
```
        fs := new(afero.SftpFs)

        fs.Connect("user", "host:port")

        fs.Mkdir("foo", os.FileMode(0000))
        fs.Mkdir("bar", os.FileMode(0777))

        f, err := fs.Create("dusse")
        if err != nil {
                log.Fatalf("open: %v", err)
        }
        defer f.Close()

        f.Write([]byte("awesomeeeeee"))
       
        fs.MkdirAll("dir1/dir2/dir3", os.FileMode(0777))

        fs.Remove("dir1/dir2/dir3")
        fs.Remove("dir1/dir2")

        fs.Disconnect()
``` 

The thing lives currently in my forked branch to have some talking before I'm creating a pull request because I felt it is not ready to be merged yet.

https://github.com/xor-gate/afero/blob/sftp/sftp.go
https://github.com/xor-gate/afero/blob/sftp/sftp/file.go

Kind regards,
Jerry

spf13:
Instead of having the connect and disconnect part of the fs, I think it would be much cleaner to pass the connected handle into the afero.SftpFs. 

See how this implementation of AferoS3 does it for inspiration. 

https://github.com/chonthu/aferoS3/blob/master/s3Fs.go

xor-gate:
@spf13 thanks for your comments. I also wasn't satisfied but it was the quickest way to get it working. I have changed it to match the s3Fs implementation and added a test file.

See https://github.com/xor-gate/afero/commit/669780de6e465b29eab862efb9704d342961da02

spf13:
Let's add some details in the readme about how to use this, call it beta, merge it in and get feedback from others?

xor-gate:
I will close this because it kind of works now (beta) and the pull request needs some discussion.

PR #48, opened 2015-12-26
Add SftpFs beta backend

xor-gate:
As discussed in #47, have fun :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,12 +263,17 @@ backed file implementation. This can be used in other memory backed file
 systems with ease. Plans are to add a radix tree memory stored file
 system using InMemoryFile.
 
+## SftpFs
+
+Afero has experimental support for secure file transfer protocol (sftp). Which can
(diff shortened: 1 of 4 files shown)


xor-gate:
It seems continues integration chokes on the test, i have no real clue how to make this actually work. The problem is we need to spawn a sftp server and test against it. Or remove the "example/test" completely.

xor-gate:
@mbertschler you are right I fixed up the test and now spawn a Sftp server and generate the RSA host key :+1:

mbertschler:
@xor-gate the Sftp server in the tests is a good idea :+1:.
I have not tested it yet but I think this is ready to merge as an experimental feature. I will probably play around with it in the next days, and also see if this FS works with our existing tests.

xor-gate:
Thank you @mbertschler. And yes having the sftp server stub in the test makes sure with the CI tests that everything is still functional.

PR #49, opened 2015-12-30
Add BasePathFs as FilterFs

vetinari:
* add a BasePathFs implemented as FilterFs:
  The BasePathFs restricts all operations to a given path within an Fs.
  The given file name to the operations on this Fs will be prepended
  with the base path before calling the base Fs.
  Any file name (after filepath.Clean()) outside this base path will be
  treated as non existing file.
  This BasePathFs will be useful when doing a disk cache inside  a given directory

* fix meaning of "facere" - "facio" means "I do", "facere" is the base
  form

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,10 @@ backed file implementation. This can be used in other memory backed file
 systems with ease. Plans are to add a radix tree memory stored file
 system using InMemoryFile.
 
+### BasePathFs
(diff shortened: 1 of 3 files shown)


spf13:
Merged as 205066d39194ad50b4daa6537f9f4d13e5a46e92

PR #50, opened 2016-01-02
add union fs

vetinari:
This adds two union fs types:
* a copy on write union where every change is done only in the overlay which is created via 
```go
  NewUnionFs(baseFs, layerFs, NewCoWUnionFs())
```
* a caching union
  - if cache time is 0, the base layer is never consulted again, once the file is in the cache
  -  with cache time > 0 the modification time of a file is used as reference

writes will be done in both layers
```go
  NewUnionFs(baseFs, layerFs, NewCacheUnionFs(cacheTime))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,15 +317,22 @@ Available UnionFs are:
 
 ### NewCacheUnionFs(time.Duration)
 
-Cache files in the layer for the given time.Duration, a cache duration of 0 means
-"forever". If the base filesystem is writeable, calls like Mkdir(), Rename() or 
-Chtimes() will be also passed down to the base layer. Writing to a file is (currently)
-not forwarded to the base.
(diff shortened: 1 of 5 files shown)


vetinari:
someone with access and knowledge on windows should check why the appveyor build https://ci.appveyor.com/project/spf13/afero/build/44 fails. From the log it looks like the modification time did not change, i.e. the second Create in
```go
    fh, _ = base.Create("/data/file.txt")
    fh.WriteString("Another test")
    time.Sleep(2 * time.Second)
    fh.Close()
```
happens at the same `time.Now()` ... see also https://github.com/spf13/afero/issues/51

spf13:
Merged as e68b257a2bacb37491373d2204329a3e7e63a3d4

PR #52, opened 2016-01-02
prevent changes when O_RDONLY is set

vetinari:


mbertschler:
I think this should have the same like f0b9fc1.
I was just confused because `os.O_RDONLY` is 0, so I just tested `if flag == os.O_RDONLY` in my code.
Thank you for the issues with the detailed test cases, I added them to the test suite.

Issue #51, opened 2016-01-02
Close()ing a MemMapFs file opened R/W should set mod time


vetinari:
Currently only files Create()d get their modification time changed, i.e. OpenFile() with flags os.O_RDRW / os.O_WRONLY never changes the modification time of a file

PR #56, opened 2016-01-04
set modtime on any write

vetinari:
* fixes https://github.com/spf13/afero/issues/51
* additionally checks for truncate on r/o files.


mbertschler:
Looks good, thank you!

PR #63, opened 2016-01-24
fix for OsFs w/ lower timestamp resolution

vetinari:
With a sleep time < 1 second, the test fails on my machine, because the underlying FS only uses timestamp resolutions of 1 second


spf13:
Merged as a5d3e7c

Issue #64, opened 2016-01-30
"go test" fails on Windows with current Afero HEAD


anthonyfok:
As of today, e.g. at commit a5d3e7c, `go test` on Afero fails these two tests on Windows:

* TestNestedDirOverlayOsFsReaddir
* TestCopyOnWriteFsWithOsFs

Console log:

```
Veronica@Veronica-V5 MINGW64 ~/GoPath/src/github.com/spf13/afero (master)
$ go test ./...
error removing tempDir C:\Users\Veronica\AppData\Local\Temp\625740957 remove C:\Users\Veronica\AppData\Local\Temp\625740957\home\test: The directory is not empty.
--- FAIL: TestNestedDirOverlayOsFsReaddir (0.01s)
        composite_test.go:45: error cleaning up tempDirs
error removing tempDir C:\Users\Veronica\AppData\Local\Temp\054838104 remove C:\Users\Veronica\AppData\Local\Temp\054838104\home: The directory is not empty.
error removing tempDir C:\Users\Veronica\AppData\Local\Temp\512058071 remove C:\Users\Veronica\AppData\Local\Temp\512058071\home: The directory is not empty.
--- FAIL: TestCopyOnWriteFsWithOsFs (0.01s)
        composite_test.go:45: error cleaning up tempDirs
FAIL
FAIL    github.com/spf13/afero  6.496s
?       github.com/spf13/afero/mem      [no test files]
?       github.com/spf13/afero/sftp     [no test files]
```

Console log in verbose test mode:

```
Veronica@Veronica-V5 MINGW64 ~/GoPath/src/github.com/spf13/afero (master)
$ go test -v ./...
=== RUN   TestRead0
--- PASS: TestRead0 (0.00s)
=== RUN   TestOpenFile
--- PASS: TestOpenFile (0.00s)
=== RUN   TestCreate
--- PASS: TestCreate (0.00s)
=== RUN   TestMemFileRead
--- PASS: TestMemFileRead (0.00s)
=== RUN   TestRename
--- PASS: TestRename (0.00s)
=== RUN   TestRemove
--- PASS: TestRemove (0.01s)
=== RUN   TestTruncate
--- PASS: TestTruncate (0.00s)
=== RUN   TestSeek
--- PASS: TestSeek (0.00s)
=== RUN   TestReadAt
--- PASS: TestReadAt (0.00s)
=== RUN   TestWriteAt
--- PASS: TestWriteAt (0.00s)
=== RUN   TestReaddirnames
[...]

PR #65, opened 2016-01-30
Fix OpenBSD build

pjanouch:
Possibly other BSDs are affected as well but this fixed the build of the whole hugo package for me.


pjanouch:
AppVeyor failure doesn't seem to be my fault.

mbertschler:
@pjanouch Thank you! Yes appveyor fail is not your fault and is already tracked in #64.

PR #75, opened 2016-03-21
Added os.Lstat to afero https://golang.org/pkg/os/#Lstat

dreh23:


spf13:
Why do we need Lstat? 
What does not having it prohibit us from being able to do?

dreh23:
From the documentation: "Lstat makes no attempt to follow the link." This would be the feature I am using in my project. Afero should be a drop-in replacement/abstraction for os, so I believe adding more calls would be useful.
For your second question I don't know the answer - is there something I don't get?

dreh23:
Maybe I can give some context. I would like to abstract os calles to test our linking tool:
I'm using os.Symlink, os.Lstat and os.Lchown in our tool. Can I replace these calls with afero or do I have to rewrite my code to fit afero? If so could you point me in the right direction - I seem to fail to figure out how to do this in an elegant way.

kofalt:
I also have need for an abstraction layer that has advanced capabilities.
In my case, I wrote a second interface and combined it with afero:

```go
import (
	"time"
	"os"

	"github.com/spf13/afero"
)

// A filesystem interface that can do a bit more than afero's (excellent & reasonable) Fs interface.
// Specifically, this adds our AdvancedFs functions which add hardlink, symlink, and similar operations.
type Fs interface {
	afero.Fs
	AdvancedFs
}

// A variety of OS functions that steal their signature & meaning from the stdlib os package.
// These are intended to be purely additive to afero's Fs interface.
//
// Please see the stdlib for documentation:
// https://golang.org/pkg/os
//
type AdvancedFs interface {
	Lchown(name string, uid, gid int) error
	Link(oldname, newname string) error
	Lstat(name string) (os.FileInfo, error)
	Mkfifo(path string, mode uint32) (err error)
	Mknod(path string, mode uint32, dev int) (err error)
	Readlink(name string) (string, error)
	Symlink(oldname, newname string) error
}
```

This lets me abstract away advanced filesystem functionality without forking, or adding each function to an interface frequently used by mock implementations with a smaller surface area. If desired, I can retool that interface into a PR for afero, or we can leave it here as advice to the few of us who need these functions.

dreh23:
Thanks Kofalt for this solution. Ideally I believe this should be integrated in afero, just so replacement is more transparent without having another abstraction layer.

bep:
A relevant note:

I have added a working branch related to this here:

https://github.com/spf13/afero/commit/c11f1d913c947c3b97a648fe91addccb0ea94e45

It is geared towards what we need in Hugo. `Lstat` is only relevant for the `os` file system, but when you combine these filesystems it is hard (or impossible) to get to the underlying FS to decide if it is a OS or something totally different. There is even a `lstatIfOs` private method inside Afero, so the use case is common enough.

PR #82, opened 2016-04-20
Revert previous fullPath changes - add tests for nested BasePaths

francoishill:
Revert the previous `fullPath` changes and add tests for "nested" `BasePath`'s

### DIFF
--- a/basepath.go
+++ b/basepath.go
@@ -42,24 +42,6 @@ func (b *BasePathFs) RealPath(name string) (path string, err error) {
 	return path, nil
 }
 
-func (b *BasePathFs) fullPath(name string) (path string, err error) {
-	if err := validateBasePathName(name); err != nil {
-		return "", err
-	}
-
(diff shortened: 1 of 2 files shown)


mbertschler:
hmm okay, but what will be the right long term solution?

francoishill:
Well the original code is actually perfect because it works as expected with "nested" BasePaths on all methods (`MkdirAll`, `Create`, etc) except the `RealPath`. So for "interacting" with the nested file system it acts on the correct/nested files/folders.

My problem is that I needed a solution to obtain the full path. Because in my case I write files to a location (using a BasePathFs that got built up in stages so they are nested). Then it must send the "full path" to an external service.

But the solution is perhaps the pull request 83 (https://github.com/spf13/afero/pull/83) I created? It adds a utility method `FullBaseFsPath` to resolve full paths of possible "nested" base paths.

What is your take on this?

PR #83, opened 2016-04-20
Add `FullBaseFsPath` utility method

francoishill:
This method resolves the "Full" path of the `BasePathFs`. It also resolves for nested BasePaths.
Unit tests were also added

### DIFF
--- a/basepath_test.go
+++ b/basepath_test.go
@@ -109,14 +109,13 @@ func TestNestedBasePaths(t *testing.T) {
 		level3Fs := NewBasePathFs(level2Fs, ds.Dir3)
 
 		type spec struct {
-			BaseFs       Fs
-			FileName     string
-			ExpectedPath string
+			BaseFs   Fs
+			FileName string
 		}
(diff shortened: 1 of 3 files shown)


mbertschler:
as a reply on:

> francoishill commented an hour ago:
> Well the original code is actually perfect because it works as expected with "nested" BasePaths on all methods (MkdirAll, Create, etc) except the RealPath. So for "interacting" with the nested file system it acts on the correct/nested files/folders.
> 
> My problem is that I needed a solution to obtain the full path. Because in my case I write files to a location (using a BasePathFs that got built up in stages so they are nested). Then it must send the "full path" to an external service.
> 
> But the solution is perhaps the pull request 83 (#83) I created? It adds a utility method FullBaseFsPath to resolve full paths of possible "nested" base paths.
> 
> What is your take on this?

The only concern would be the security concern that was mentioned before, about not revealing paths to the user, but I could not come up with an idea where this would be relevant.

So I welcome this new function.

PR #85, opened 2016-04-23
Fix test and add `CopyFile` and `CopyDir` util funcs

francoishill:


util.go
```diff
@@ -329,3 +329,76 @@ func FullBaseFsPath(basePathFs *BasePathFs, relativePath string) string {
...
+		return err
+	}
+	defer destFile.Close()
+
+	_, err = io.Copy(destFile, srcFile)
+	if err != nil {
+		return err
+	}
+
+	if err != nil {
```
xuzhenglun: this err has been handled in line 353. 

maybe below is better?
```suggestion
if err := destFs.Chmod(destDirPath, srcInfo.Mode()); err != nil {
        return err
}
```

util.go
```diff
@@ -329,3 +329,76 @@ func FullBaseFsPath(basePathFs *BasePathFs, relativePath string) string {
...
+		return err
+	}
+
+	directory, err := srcFs.Open(srcDirPath)
+	if err != nil {
+		return err
+	}
+	defer directory.Close()
+
+	entries, err := directory.Readdir(-1)
```
xuzhenglun: do we need to handle this error?

util.go
```diff
@@ -329,3 +329,76 @@ func FullBaseFsPath(basePathFs *BasePathFs, relativePath string) string {
...
+	// Some code from https://www.socketloop.com/tutorials/golang-copy-directory-including-sub-directories-files
+
+	// get properties of source dir
+	srcInfo, err := srcFs.Stat(srcDirPath)
+	if err != nil {
+		return err
+	}
+
+	// create dest dir
+	if err = destFs.MkdirAll(destDirPath, srcInfo.Mode()); err != nil {
```
xuzhenglun: maybe we should set dir mode after all contents of it are copied.

kargakis:
Closing due to conflicts, please feel free to reopen if you wish to fix and merge.

PR #87, opened 2016-04-28
test copyOnWrite Stat()

pgruenbacher:
So this next issue with copyOnWrite was that the when opening a file that existed in the os but not in memory, the Stat function error was returning a os.NotExist error but wasn't getting caught appropriately.
It may warrant further exploration why the memory layer was returning an os.NotExistErr

the test will fail if the `if err == syscall.ENOENT || err == syscall.ENOTDIR {` is left in


bep:
Fixed in 8d919cbe7e2627e417f3e45c3c0e489a5b7e2536

PR #90, opened 2016-05-22
S3 FS implementation

rgarcia:
This is a mostly-working/mostly-tested implementation of S3 as an FS.

There are some gaps in the implementation, but it is otherwise pretty functional:
- It doesn't support non-offset == 0 reads or writes
- It doesn't support Truncate
- It doesn't support `Seek` with whence set to 2
- It doesn't support opening a file in append mode

All of these things increase the complexity of the implementation significantly (some are questionably well-defined, e.g. append in an eventually-consistent FS like S3), so for now I've left them out.

In order to reduce clutter in the root package `github.com/spf13/afero`, I put the whole implementation + tests in a sub package, `github.com/spf13/afero/s3`. In order to leverage the generic `Fs` tests in `afero_test.go`, I copy-pasted them into a sub package `github.com/spf13/afero/tests`. I also added a few tests to this file as I went along.

In order to run the tests you must pass a bucket: `go test -v github.com/spf13/afero/s3 -s3bucket <test bucket>`. You also need to set `AWS_REGION`, `AWS_ACCESS_KEY_ID`, and `AWS_SECRET_ACCESS_KEY`.

Let me know what you think 😄

### DIFF
--- a/path.go
+++ b/path.go
@@ -24,7 +24,7 @@ import (
 // readDirNames reads the directory named by dirname and returns
 // a sorted list of directory entries.
 // adapted from https://golang.org/src/path/filepath/path.go
-func readDirNames(fs Fs, dirname string) ([]string, error) {
+func ReadDirNames(fs Fs, dirname string) ([]string, error) {
 	f, err := fs.Open(dirname)
 	if err != nil {
 		return nil, err
@@ -53,7 +53,7 @@ func walk(fs Fs, path string, info os.FileInfo, walkFn filepath.WalkFunc) error 
 		return nil
 	}
 
-	names, err := readDirNames(fs, path)
+	names, err := ReadDirNames(fs, path)
 	if err != nil {
 		return walkFn(path, info, err)
 	}
--- a/s3/s3_file.go
+++ b/s3/s3_file.go
@@ -1,0 +1,266 @@
+package s3
+
+import (
+	"bytes"
+	"io"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3iface"
+	"github.com/spf13/afero"
+)
+
+// S3File represents a file in S3.
+// It is not threadsafe.
+type S3File struct {
+	bucket string
+	name   string
+	s3Fs   afero.Fs
+	s3API  s3iface.S3API
+
+	// state
+	offset int
+	closed bool
+
+	// readdir state
+	readdirContinuationToken *string
+	readdirNotTruncated      bool
+}
+
+var _ afero.File = (*S3File)(nil)
+
+// NewS3File initializes an S3File object.
+func NewS3File(bucket, name string, s3API s3iface.S3API, s3Fs afero.Fs) *S3File {
+	return &S3File{
+		bucket: bucket,
+		name:   name,
+		s3API:  s3API,
+		s3Fs:   s3Fs,
+		offset: 0,
+		closed: false,
+	}
+}
+
+// Name returns the filename, i.e. S3 path without the bucket name.
+func (f *S3File) Name() string { return f.name }
+
+// Readdir reads the contents of the directory associated with file and
+// returns a slice of up to n FileInfo values, as would be returned
+// by ListObjects, in directory order. Subsequent calls on the same file will yield further FileInfos.
+//
+// If n > 0, Readdir returns at most n FileInfo structures. In this case, if
+// Readdir returns an empty slice, it will return a non-nil error
+// explaining why. At the end of a directory, the error is io.EOF.
+//
+// If n <= 0, Readdir returns all the FileInfo from the directory in
+// a single slice. In this case, if Readdir succeeds (reads all
+// the way to the end of the directory), it returns the slice and a
+// nil error. If it encounters an error before the end of the
+// directory, Readdir returns the FileInfo read until that point
+// and a non-nil error.
+func (f *S3File) Readdir(n int) ([]os.FileInfo, error) {
+	if f.readdirNotTruncated {
+		return nil, io.EOF
+	}
+	if n <= 0 {
+		return f.ReaddirAll()
+	}
+	// ListObjects treats leading slashes as part of the directory name
+	// It also needs a trailing slash to list contents of a directory.
+	name := trimLeadingSlash(f.Name()) + "/"
+	output, err := f.s3API.ListObjectsV2(&s3.ListObjectsV2Input{
+		ContinuationToken: f.readdirContinuationToken,
+		Bucket:            aws.String(f.bucket),
+		Prefix:            aws.String(name),
+		Delimiter:         aws.String("/"),
+		MaxKeys:           aws.Int64(int64(n)),
+	})
+	if err != nil {
+		return nil, err
+	}
+	f.readdirContinuationToken = output.NextContinuationToken
+	if !(*output.IsTruncated) {
+		f.readdirNotTruncated = true
+	}
+	fis := []os.FileInfo{}
+	for _, subfolder := range output.CommonPrefixes {
+		fis = append(fis, NewS3FileInfo(filepath.Base("/"+*subfolder.Prefix), true, 0, time.Time{}))
+	}
+	for _, fileObject := range output.Contents {
+		if hasTrailingSlash(*fileObject.Key) {
+			// S3 includes <name>/ in the Contents listing for <name>
+			continue
+		}
+
+		fis = append(fis, NewS3FileInfo(filepath.Base("/"+*fileObject.Key), false, *fileObject.Size, *fileObject.LastModified))
+	}
+
+	return fis, nil
+}
+
+func (f *S3File) ReaddirAll() ([]os.FileInfo, error) {
+	fileInfos := []os.FileInfo{}
+	for {
+		infos, err := f.Readdir(100)
+		fileInfos = append(fileInfos, infos...)
+		if err != nil {
+			if err == io.EOF {
+				break
+			} else {
+				return nil, err
+			}
+		}
+	}
+	return fileInfos, nil
+}
+
+// Readdirnames reads and returns a slice of names from the directory f.
+//
+// If n > 0, Readdirnames returns at most n names. In this case, if
+// Readdirnames returns an empty slice, it will return a non-nil error
+// explaining why. At the end of a directory, the error is io.EOF.
+//
+// If n <= 0, Readdirnames returns all the names from the directory in
+// a single slice. In this case, if Readdirnames succeeds (reads all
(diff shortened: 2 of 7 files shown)


dim:
:+1:

ns-cweber:
Could this get some love? I am very interested in using this in my project. Depending on a fork is never ideal.

s3/s3_fs.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// Open a file for reading.
+// If the file doesn't exist, Open will create the file.
+func (fs S3Fs) Open(name string) (afero.File, error) {
+	if _, err := fs.Stat(name); err != nil {
+		if os.IsNotExist(err) {
+			return fs.OpenFile(name, os.O_CREATE, 0777)
+		}
+		return (*S3File)(nil), err
```
ns-cweber: Any reason you're returning a valid interface to a nil pointer? While everyone should be checking the error, CacheOnReadFs checks the interface for nil, and this breaks in hard-to-troubleshoot ways. While I don't think anyone should be ignoring the error, returning a valid interface isn't doing anyone any favors either.

s3/s3_fs.go
```diff
@@ -0,0 +1,235 @@
...
+	nameClean := filepath.Clean(name)
+	out, err := fs.s3API.HeadObject(&s3.HeadObjectInput{
+		Bucket: aws.String(fs.bucket),
+		Key:    aws.String(nameClean),
+	})
+	if err != nil {
+		if strings.Contains(err.Error(), "404") {
+			statDir, err := fs.statDirectory(name)
+			return statDir, err
+		}
```
ns-cweber: Rather than munging the string, I think it would be better to check the error code:

```go
if err, ok := err.(awserr.Error); ok {
    if err.Code() == "NotFound" {
        return fs.statDirectory(name)
    }
}
```

ns-cweber: Here's a longer, runnable example to play around with (just manipulate the bucket name):

```go
package main

import (
	"log"

	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/awserr"
	"github.com/aws/aws-sdk-go/aws/session"
	"github.com/aws/aws-sdk-go/service/s3"
)

func main() {
	sess, err := session.NewSession()
	if err != nil {
		fmt.Fprintln(os.Stderr, err)
		os.Exit(-1)
	}

	if _, err := s3.New(sess).HeadObject(&s3.HeadObjectInput{
		Bucket: aws.String("bucketName"),
		Key:    aws.String("/missing/key/"),
	}); err != nil {
		if err, ok := err.(awserr.Error); ok {
			fmt.Println("CODE:", err.Code())
		}
	}
```

s3/s3_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/spf13/afero/test"
+)
+
+var s3bucket string
+
+func init() {
```
nicored: This should be done as an integration test. Instead, you could Money Patch s3 methods (mock cannot really be done).

```go
// my_test.go
...
monkey.PatchInstanceMethod(reflect.TypeOf(s3Session), "GetObject", s3_mocks.GetObjectMockFn())
...
```

```go
// mocks_s3.go

type ClosingBuffer struct {
    *bytes.Buffer
}

func (cb *ClosingBuffer) Close() error {
	return nil
}

func GetObjectMockFn() func(_ *s3.S3, _ *s3.GetObjectInput) (*s3.GetObjectOutput, error) {
	return func(_ *s3.S3, _ *s3.GetObjectInput) (*s3.GetObjectOutput, error) {
		cb := &ClosingBuffer{bytes.NewBufferString(`This is the content of a file`)}

		var rc io.ReadCloser
		rc = cb
		rc.Close()

		return &s3.GetObjectOutput{
			Body: rc,
		}, nil
	}
}

Comment (justinclift) [inline s3/s3_test.go]: Maybe Minio as a semi-functional mock?

Comment (nicored) [inline s3/s3_test.go]: Which would mean that your unit tests would depend on Minio, which I'm against. 
Your unit tests should not depend on external services to work, but Minio would be a good candidate for integration testing, but then I'd rather connect my tests to a test bucket in s3.
I've had a chat with a few people who run their unit tests as integration tests to avoid to have to re-write parts of their tests, which sounds wrong to me.
Here, we're using the aws sdk, which we assume is stable and has been tested, so what you want to is just test your app logic and mock what functions will return, and allow you to also test how you handle errors you'd not get when aws behaves fine. It gives you more flexibility in your unit tests to catch all edge cases you can think of. I do agree that writing mocks can often add complexity in your tests, but it's worth it.

Comment (justinclift) [inline s3/s3_test.go]: Yeah, I can see where you're coming from.  Using S3 itself as an integration test sounds potentially slow (and $) too, but I guess that depends on a lot of factors.

With mock vs integration... neither way seems to be perfect, both being needed. :wink:

Comment (nicored) [inline s3/s3_test.go]: Arguably, storage is really really cheap these days, 0.0245 per gb for standard storage.

**Comment by jcftang** [comment]
I know this has been sitting around for a while, but it would be nice to see if one of the s3 backends could be merged in

**Comment by aviau** [comment]
Any news on this? I would also love to see this merged.

**Comment by aviau** [comment]
Someone should probably move this PR to https://github.com/absfs/afero

File: path.go [hunk 4]
```diff
@@ -24,7 +24,7 @@ import (
 // readDirNames reads the directory named by dirname and returns
```
kargakis: Can you also update the function name in the comment?

s3/s3_file.go
```diff
@@ -0,0 +1,266 @@
...
+// Read reads up to len(b) bytes from the File.
+// It returns the number of bytes read and an error, if any.
+// EOF is signaled by a zero count with err set to io.EOF.
+func (f *S3File) Read(p []byte) (int, error) {
+	if f.closed {
+		// mimic os.File's read after close behavior
+		panic("read after close")
+	}
+	if f.offset != 0 {
+		panic("TODO: non-offset == 0 read")
```
kargakis: Maybe open a follow/up issue to track this.

s3/s3_file.go
```diff
@@ -0,0 +1,266 @@
...
+// It returns the new offset and an error, if any.
+// The behavior of Seek on a file opened with O_APPEND is not specified.
+func (f *S3File) Seek(offset int64, whence int) (int64, error) {
+	switch whence {
+	case 0:
+		f.offset = int(offset)
+	case 1:
+		f.offset += int(offset)
+	case 2:
+		// can probably do this if we had GetObjectOutput (ContentLength)
```
kargakis: Seems to be a thing now

https://godoc.org/github.com/aws/aws-sdk-go/service/s3#GetObjectOutput

s3/s3_file.go
```diff
@@ -0,0 +1,266 @@
...
+// Write writes len(b) bytes to the File.
+// It returns the number of bytes written and an error, if any.
+// Write returns a non-nil error when n != len(b).
+func (f *S3File) Write(p []byte) (int, error) {
+	if f.closed {
+		// mimic os.File's write after close behavior
+		panic("write after close")
+	}
+	if f.offset != 0 {
+		panic("TODO: non-offset == 0 write")
```
kargakis: This seems like it should be fixed, otherwise Write in existing files is not useful?

s3/s3_file.go
kargakis: On a second thought, this should probably be fixed from the get-go.

s3/s3_file.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	readSeeker := bytes.NewReader(p)
+	size := int(readSeeker.Size())
+	if _, err := f.s3API.PutObject(&s3.PutObjectInput{
+		Bucket:               aws.String(f.bucket),
+		Key:                  aws.String(f.name),
+		Body:                 readSeeker,
+		ServerSideEncryption: aws.String("AES256"),
```
kargakis: For my own info. how is the key managed server-side? Is it fully managed by Amazon?

s3/s3_fs.go
```diff
@@ -0,0 +1,235 @@
...
+	// Create(), like all of S3, is eventually consistent.
+	// To protect against unexpected behavior, have this method
+	// wait until S3 reports the object exists.
+	if s3Client, ok := fs.s3API.(*s3.S3); ok {
+		return file, s3Client.WaitUntilObjectExists(&s3.HeadObjectInput{
+			Bucket: aws.String(fs.bucket),
+			Key:    aws.String(name),
+		})
+	}
+	return file, err
```
kargakis: nit:
```go
return file, nil
```

s3/s3_fs.go
```diff
@@ -0,0 +1,235 @@
...
+// Stat returns a FileInfo describing the named file.
+// If there is an error, it will be of type *os.PathError.
+func (fs S3Fs) Stat(name string) (os.FileInfo, error) {
+	nameClean := filepath.Clean(name)
+	out, err := fs.s3API.HeadObject(&s3.HeadObjectInput{
+		Bucket: aws.String(fs.bucket),
+		Key:    aws.String(nameClean),
+	})
+	if err != nil {
+		if strings.Contains(err.Error(), "404") {
```
kargakis: awserr is a thing so please try to cast to typed errors

https://docs.aws.amazon.com/sdk-for-go/api/aws/awserr/

kargakis:
@rgarcia if you have time addressing the outstanding comments, I can help with merging.

rgarcia:
Thanks for the review @kargakis and hi everyone. I made this PR almost 4 years ago and since then my team's priorities have shifted and we no longer use afero. So unfortunately it's very unlikely that I will find the time to finish the work here.

dim:
Thanks @rgarcia, that's why we have come up with our own spin on afero https://github.com/bsm/bfs, it has adapters for in-memory, file system, S3, GCS, FTP and soon AZ Blob

fclairamb:
Hi guys,

I reworked a fork of this implementation [here](https://github.com/fclairamb/afero-s3) and I'm interested by your feedback on it. It's based on @wreulicke's fork of @aviau's fork of @rgarcia's code and basically fixes some limitations it had. The biggest limitation I had is that it didn't handle the streaming (for upload or download) of files. I used a pipe and a go-routine to handle that.

In my humble opinion:
- Each Fs implementation should be in its own repository so that each developer can choose the one he wants
- You should recommend some of them or even create an `afero-contrib` github organization where you share the core Fs

I intend to build some other Afero.Fs for a multi-backend [ftp server](https://github.com/fclairamb/ftpserver) which was initially only [a library](https://github.com/fclairamb/ftpserverlib).

kargakis:
Hi Florent,

if you want to submit your S3 backend as a PR to this repo, I'll happily help with reviewing and merging your work.

-------- Original Message --------

pinpox:
Is there still work being done on this?

PR #98, opened 2016-09-19
Added glob/match support to afero

kcrawley:
I needed the [`Glob`](https://golang.org/pkg/path/filepath/#Glob) functionality that Golang supports in their `path/filepath` package. I suppose someone else could use this, perhaps.


kcrawley:
any interest in this?

Issue #89, opened 2016-05-18
move fs implementations into sub packages


rgarcia:
Right now if I depend on `github.com/spf13/afero`, I also have to pull in a lot of potentially unnecessary dependencies that might not be related to the FS implementation I'm using, e.g. `github.com/pkg/sftp`.

Let me know if you agree and would be open to a PR.

kofalt:
For reference, here's the current set of remote dependencies, do you know which would be trimmed?

```
$ go get -v github.com/spf13/afero

github.com/spf13/afero (download)
github.com/pkg/sftp (download)
github.com/kr/fs (download)
github.com/pkg/errors (download)
golang.org/x/crypto (download)
golang.org/x/text (download)
```

Godoc also has the [imports](https://godoc.org/github.com/spf13/afero?imports) and [graph](https://godoc.org/github.com/spf13/afero?import-graph).

rgarcia:
The graph with std lib hidden is informative: https://godoc.org/github.com/spf13/afero?import-graph&hide=1

It looks like almost all of the third party deps fall from the `sftp` FS. This suggests the lowest hanging fruit might be to isolate the `sftp` implementation in the `github.com/afero/sftp` package.

kofalt:
Works for me! I definitely support sub-packaging when implementations have unique import requirements.
I'm also not the maintainer, though :smile:

PR #104, opened 2016-11-03
Move sftpfs to subpackage

### LINKED ISSUE #97: sftp should live in its subpackage only
Right now it is a confusing mess -- and you end up downloading sftp dependencies even if you don't use sftp.

moorereason:
Fixes #97


moorereason:
Updates #89

PR #119, opened 2017-03-24
Remove syscall package dependency for AppEngine

CaptainCodeman:
Moves dependencies into a separate const.go file that can be build tagged out.

An alternative, non syscall referencing file can then be used to make the package usable on appengine (with https://github.com/captaincodeman/afero-datastore)

### DIFF
--- a/afero_test.go
+++ b/afero_test.go
@@ -23,7 +23,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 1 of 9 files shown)


thefallentree:
👍

CaptainCodeman:
Probably not required now the second generation Go runtime 1.11 is available.

PR #120, opened 2017-03-25
readme: remove inferred type from AppFs declaration.

thylong:
AppFs type is inferred from afero.NewOsFs(), go lint tools are complaining about expliciting the type.

> should omit type afero.Fs from declaration of var AppFs; it will be inferred from the right-hand side (golint)


thefallentree:
👍

Issue #115, opened 2017-02-18
CacheOnReadFs.Stat() returns FileNotFound when file exists on base but not on cache


ns-cweber:
I'm on OS X 10.11.6

Given directories `/tmp/base/` and `/tmp/cache` wherein the former contains a file `foo` and the latter is empty, the following code fails with the error message: `CACHE_ON_READ: stat /tmp/cache/foo: no such file or directory`. Because `foo` exists in `base`, I was expecting to see a valid FileInfo and no error.

```go
func main() {
	base := afero.NewBasePathFs(afero.NewOsFs(), "/tmp/base/")
	if _, err := base.Stat("foo"); err != nil {
		log.Fatalln("BASE:", err)
	}
	fs := afero.NewCacheOnReadFs(
		base,
		afero.NewBasePathFs(afero.NewOsFs(), "/tmp/cache/"),
		time.Duration(0),
	)
	if _, err := fs.Stat("foo"); err != nil {
		log.Fatalln("CACHE_ON_READ:", err)
	}
}
```

uppfinnarn:
Trivially reproducible with `MemMapFs` instances. This is a slight modification of the test case in the README:

```go
func TestCacheOnReadFs(t *testing.T) {
	appFS := afero.NewOsFs()
	// create test files and directories
	appFS.MkdirAll("src/a", 0755)
	afero.WriteFile(appFS, "src/a/b", []byte("file b"), 0644)
	afero.WriteFile(appFS, "src/c", []byte("file c"), 0644)
	name := "src/c"
	cacheFS := afero.NewCacheOnReadFs(appFS, afero.NewMemMapFs(), 1*time.Hour)
	_, err := cacheFS.Stat(name)
	if os.IsNotExist(err) {
		t.Errorf("file \"%s\" does not exist.\n", name)
	}
}
```

mbertschler:
Fixed via #135

PR #121, opened 2017-03-25
Fix failure of the cache miss processing

aspcartman:
Error was not properly checked. I guess it's a simple typo. As the result instead of processing the error as the special cache-miss case, it was returning the underlying error up to the caller.

### DIFF
--- a/cacheOnReadFs.go
+++ b/cacheOnReadFs.go
@@ -67,9 +67,8 @@ func (u *CacheOnReadFs) cacheStatus(name string) (state cacheState, fi os.FileIn
 	if err == syscall.ENOENT {
 		return cacheMiss, nil, nil
 	}
-	var ok bool
-	if err, ok = err.(*os.PathError); ok {
-		if err == os.ErrNotExist {
+	if err, ok := err.(*os.PathError); ok {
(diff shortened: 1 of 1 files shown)


aspcartman:
Fixes #115

thefallentree:
👍

cacheOnReadFs.go
```diff
@@ -67,9 +67,8 @@ func (u *CacheOnReadFs) cacheStatus(name string) (state cacheState, fi os.FileIn
 	if err == syscall.ENOENT {
 		return cacheMiss, nil, nil
 	}
-	var ok bool
-	if err, ok = err.(*os.PathError); ok {
-		if err == os.ErrNotExist {
```
moorereason: Looks like you're in the right place, but this could use a couple tweaks.  I would remove the `if err == syscall.ENOENT {}` above and change your patch to:

```go
if err, ok := err.(*os.PathError); ok {
    if os.IsNotExist(err) {
```

`os.IsNotExist` will take care of syscall checks in a cross-platform way (`ENOENT` is unix-only).

mbertschler:
As far as I can see this was fixed by #135.

Issue #122, opened 2017-05-08
os.Chown implementation


giulianvivan:
Any chance to implement the os.Chown in the afero filesystems?

kargakis:
Fixed in https://github.com/spf13/afero/pull/279

PR #130, opened 2017-08-17
Here I go again on my Chown

cfunkhouser:
Add a Chown function matching the os.Chown signature to afero.Fs, and implement it in the various Fs implementations.

### DIFF
--- a/mem/file.go
+++ b/mem/file.go
@@ -57,6 +57,8 @@ type FileData struct {
 	dir     bool
 	mode    os.FileMode
 	modtime time.Time
+	uid     int
+	gid     int
 }
 
 func (d *FileData) Name() string {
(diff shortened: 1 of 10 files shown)


cfunkhouser:
Fixes #122

corentone:
@cfunkhouser your timing is perfect, I'm also looking for this!
I'm not familiar with the internal of afero but I'd be happy to help.

mem/file.go
```diff
@@ -81,6 +83,14 @@ func SetMode(f *FileData, mode os.FileMode) {
 	f.mode = mode
 }
 
+func SetUID(f *FileData, uid int) {
+	f.uid = uid
```
mbertschler: This and the SetGID function will need locks, like they were recently added to master to not have races on the FileData struct.

memmap.go
```diff
@@ -347,6 +347,24 @@ func (m *MemMapFs) Chtimes(name string, atime time.Time, mtime time.Time) error
...
+	name = normalizePath(name)
+
+	m.mu.RLock()
+	f, ok := m.getData()[name]
+	m.mu.RUnlock()
+	if !ok {
+		return &os.PathError{Op: "chown", Path: name, Err: ErrFileNotFound}
+	}
+
+	m.mu.Lock()
```
mbertschler: If the SetUID and SetGID functions are using locks, we won't need this lock anymore.

kargakis:
Needs a rebase

kargakis:
Feel free to reopen if you plan to fix.

Issue #133, opened 2017-08-25
Races on the FileData struct?


corentone:
Hello, 

I'm getting a raceCondition between Read and Write to a file on the MemFS, see the snippet below:
```
WARNING: DATA RACE
Read at 0x00c42009a318 by goroutine 7:
 SOMEPATH/vendor/github.com/spf13/afero/mem.(*FileInfo).Size()
      SOMEPATH/vendor/github.com/spf13/afero/mem/file.go:275 +0x9d
  SOMEPATH/vendor/github.com/spf13/afero.ReadFile()
      SOMEPATH/vendor/github.com/spf13/afero/ioutil.go:76 +0x1bf
  SOMEPATH.readFiles()

Previous write at 0x00c42009a318 by goroutine 6:
  SOMEPATH/vendor/github.com/spf13/afero/mem.(*File).Write()
      SOMEPATH/vendor/github.com/spf13/afero/mem/file.go:236 +0x98c
  SOMEPATH/vendor/github.com/spf13/afero.WriteFile()
     SOMEPATH/vendor/github.com/spf13/afero/ioutil.go:129 +0xcb
  SOMEPATH.writeFile()
```

(I do have a routine reading the file and one writing it).

Looking deeper at the code, I'm slightly confused by the locking code in https://github.com/spf13/afero/blob/master/mem/file.go

Reading the name on *FileData is being locked (https://github.com/spf13/afero/blob/master/mem/file.go#L62) but writing it is not (https://github.com/spf13/afero/blob/master/mem/file.go#L76).

I feel a whole lots of locking are missing here or I could be missing the point of the lock?!

jszwec:
I believe that my pull request fixes this. Unfortunately It has been opened for a while now.
https://github.com/spf13/afero/pull/128

corentone:
@jszwec we might actually need both #128 and #134? If one is accessing any FileInfo directly to change the file then it would also trigger. Now we would need to make sure we don't double lock.

I like that you've added a test too. I should do that too. Let me try to ping a recent contributor if we can move forward on that.

corentone:
@bep @mbertschler (you are the two most recent merging members for PRs so pinging you), could you weight on this? It would be awesome to get those races fixed.

mbertschler:
Fixed by #134

PR #134, opened 2017-08-25
mem/file.go - Fix some races in accessing fields of FileData

corentone:
* Splitting SetModeTime to avoid double locking
* Adding locks all over the place.

Potential fix for #133

### DIFF
--- a/mem/file.go
+++ b/mem/file.go
@@ -74,14 +74,24 @@ func CreateDir(name string) *FileData {
 }
 
 func ChangeFileName(f *FileData, newname string) {
+	f.Lock()
 	f.name = newname
+	f.Unlock()
 }
 
 func SetMode(f *FileData, mode os.FileMode) {
+	f.Lock()
 	f.mode = mode
+	f.Unlock()
 }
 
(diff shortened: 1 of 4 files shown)


mem/file.go
```diff
@@ -74,14 +74,24 @@ func CreateDir(name string) *FileData {
...
 
 func ChangeFileName(f *FileData, newname string) {
+	f.Lock()
+	defer f.Unlock()
 	f.name = newname
 }
 
 func SetMode(f *FileData, mode os.FileMode) {
+	f.Lock()
+	defer f.Unlock()
```
bep: No need to use `defer` when the method is so short.

bep: Also, it would be great if you added a test that demonstrates this problem.

corentone:
@bep thanks for the review and sorry for the delay addressing them.
Let me know if you have any comment, I also added some UT for the funcs I changed which are kind of boringly verbose...

mbertschler:
I verified that the new tests fail with data races without this patch.

mbertschler:
Fixes #133

PR #135, opened 2017-09-05
CacheOnReadFS: erroneous NotExists error from MemMapFS layer

rodaine:
CacheOnReadFS erroneously returns a NotExists error even if the file is available in the base FS. This first commit demonstrates this error with a test. The following will include a fix.


mbertschler:
Thank you for this PR!

PR #137, opened 2017-10-03
Always return the concrete type instead of the Fs interface

### LINKED ISSUE #76: NewBasePathFs should return the concrete type
Having to do:

```
 NewBasePathFs(fs, baseDir).(*BasePathFs)
```

Doesn't make much sense. I don't see the problem going the other way around; but as this is a file system with a very specific contract, I want to know what it is.

mbertschler:
Added `var _ afero.Fs = &Fs{}` style type checks to statically check if the interface is implemented.

These changes broke one test:
```diff
-	bp := NewBasePathFs(fs, baseDir).(*BasePathFs)
+	bp := NewBasePathFs(fs, baseDir)
```
This pattern is probably also found in code that uses this library, which makes this a breaking change. Details are discussed in #76.

fixes #76

### DIFF
--- a/basepath.go
+++ b/basepath.go
@@ -22,7 +22,10 @@ type BasePathFs struct {
 	path   string
 }
 
-func NewBasePathFs(source Fs, path string) Fs {
+// Check if BasePathFs implements Fs
+var _ Fs = &BasePathFs{}
+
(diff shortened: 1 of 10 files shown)


kargakis:
Needs a rebase

kargakis:
Feel free to reopen if you plan to fix.

PR #139, opened 2017-10-08
Minor improvement

snwfdhmp:
It seems better and more understandable when using the same "styling" process for those 2 things (before/after)


mbertschler:
Thanks, it looks nicer that way.

PR #142, opened 2017-10-23
Use more restrictive interface for SFTP

paultyng:
This uses a more restrictive interface for the SFTP client and removes the requirement that you use `github.com/pkg/sftp` directly, which is useful if you want to introduce SFTP middleware or use a different package as a client.

Once we do this though, the coupling is pretty loose, you could probably make this more generic and get rid of the fact that its SFTP at all, its basically just a provider without the ability to `ReaderAt` or `WriterAt` I think.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,9 @@ sudo: false
 language: go
 
 go:
-  - 1.7.5
-  - 1.8
+  - 1.7.x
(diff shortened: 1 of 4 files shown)


kargakis:
Needs a rebase, feel free to reopen if you plan to fix.

PR #146, opened 2017-11-15
Add zipfs, an archive/zip-based read-only filesystem

hillu:


jcftang:
Would love to see this merged as well, I was about to start on a zipfs backend myself till I saw this.

hillu:
@bep Are you interested in this PR? If not, it is fine, in that case I would simply publish the zipfs package in a separate repository.

aviau:
@hillu we might be interested to merge this in https://github.com/absfs/afero

bep:
I will look more at this in a few days (please bump if I don't). Can I ask, what is your current use case for this?

/cc @spf13

hillu:
@bep The main (public) use case is contained in <https://github.com/DCSO/spyre>. The tool can either read its configuration from a ZIP file that has been appended to the executable or from files that have been placed into the same directory as the executable.

josephlewis42:
@bep another bump for this. Our use-case is something similar, except the zip files are embedded in the executable for legal compliance.

For us, the killer feature Afero provides is support for union filesystems, allowing users to modify/override the embedded files if they desired.

senghoo:
This package seems panic if check stat of root. can u fix it? @hillu 

`
f, err := fs.Open("/") // this err is not nil. return a File without zipfile field.
f.Stat() // panic on nil access.
`

hillu:
@senghoo done.

zipfs/file.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	if offset > int64(f.zipfile.UncompressedSize64) {
+		offset = int64(f.zipfile.UncompressedSize64)
+		err = io.EOF
+	}
+	if len(f.buf) >= int(offset) {
+		return
+	}
+	buf := make([]byte, int(offset)-len(f.buf))
+	n, _ := io.ReadFull(f.reader, buf)
```
kargakis: Do not ignore the error

zipfs/file.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	if f.closed {
+		return 0, afero.ErrFileClosed
+	}
+	switch whence {
+	case os.SEEK_SET:
+	case os.SEEK_CUR:
+		offset += f.offset
+	case os.SEEK_END:
+		offset += int64(f.zipfile.UncompressedSize64)
```
kargakis: Wont this cause an out of range error below, eg. if I want to move the offset 2 bytes before the end? Shouldn't this be:
```go
offset = int64(f.zipfile.UncompressedSize64) - offset
```
Or is offset meant to be provided as a negative number?

zipfs/fs.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/spf13/afero"
+)
+
+type Fs struct {
+	r     *zip.Reader
+	files map[string]map[string]*zip.File
+}
+
+func splitpath(name string) (dir, file string) {
```
kargakis: This is used in a couple of places and since it is a self-contained helper, it would be nice to add a couple of unit tests for it.

zipfs/file.go
hillu: > Or is offset meant to be provided as a negative number?
Exactly. See the _lseek(2)_ manpage on any Unix system.

zipfs/file.go
hillu: Fixed.

zipfs/fs.go
hillu: From my point of view this might as well happen post-merge. If this PR ever gets merged, that is.

hillu:
Thanks!

PR #147, opened 2017-11-29
fix File.Name() on BasePathFS (wrap File objects returned by BasePathFs)

arbelt:
Files returned by the BasePathFs operations don't report a usable name (using `file.Name()`) since they include the prefix, which is prepended a second time by the Fs when passed back. For instance, the following doesn't work:

```go
osfs := afero.NewOsFs()
bfs := afero.NewBasePathFs(osfs, "/some/base/path")
f, _ := bfs.Create("relative/path")
// do some stuff

// This doesn't work since it tries to remove 
// /some/base/path/some/base/path/relative/path
bfs.Remove(f.Name())
```

This is especially irksome when working with TempFiles; basically anything that uses `file.Name()` ends up needing a type switch to get a path usable by the `Fs` object, which breaks the transparency of the Fs abstraction.

I've implemented a simple fix for this: add a `BasePathFile` type with an overridden `Name` method to strip the base path. Not sure if there are more subtle consequences to this, but it seems to work so I figured I'd submit a PR.

### DIFF
--- a/basepath_test.go
+++ b/basepath_test.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"testing"
+	"fmt"
 )
 
 func TestBasePath(t *testing.T) {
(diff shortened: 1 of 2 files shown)


ts2909:
I really should check the issues before making changes myself :) I noticed the same issue and started fixing over at https://github.com/ts2909/afero/commit/6c031db617e46c3d6b8415e9f64780339555a288

The only thing I did that you haven't is some tests. Feel free to use mine if you think they are ok.

arbelt:
Thanks! I've added the tests, basically verbatim.

The automated tests also caught some issues with Windows paths. I think it's resolved now by cleaning before stripping the basepath.

basepath_test.go
```diff
@@ -140,3 +141,53 @@ func TestNestedBasePaths(t *testing.T) {
...
+func TestBasePathTempFile(t *testing.T) {
+	baseFs := &MemMapFs{}
+	baseFs.MkdirAll("/base/path/tmp", 0777)
+	bp := NewBasePathFs(baseFs, "/base/path")
+
+	tDir, err := TempDir(bp, "/tmp", "")
+	if err != nil {
+		t.Fatalf("Failed to TempDir: %v", err)
+	}
+	fmt.Println(tDir)
```
ts2909: debugging output left in the test

arbelt: my bad, just saw this! removed the println.

noonien:
Any chance to get this merged?

LasTshaMAN:
What are you waiting for ? :)
Merge it already!

mbertschler:
thanks for the fix and the tests!

PR #157, opened 2018-03-09
sftpfs: Fix test with latest API of the sftp package

xor-gate:
This fixes the disabled test with latest API from the sftp package. It is still not a very clean test but it works.


xor-gate:
Thanks @kargakis, nice to see the project moving again!

PR #158, opened 2018-03-19
sftpfs: Implement OpenFile method

kklin:
Before, the OpenFile method was unimplemented, and just returned `nil,
nil`. Because the Afero.TempFile method calls the OpenFile method, this
behavior caused a nil pointer exception when TempFile was used with
sftpfs.

This commit adds a simple implementation for sftpfs.OpenFile, fixing the
exception thrown when using it with TempFile.


kargakis:
failure is from stale go1.8 build in travis, tip is passing, merging

PR #159, opened 2018-03-23
WORK IN PROGRESS: Add an optional Lstater interface

bep:
The interface has one method returning the `FileInfo` and a flag telling if `Lstat` was called or not.

```go
type Lstater interface {
    LstatIfPossible(name string) (os.FileInfo, bool, error)
}
```

`Lstat` is currently only supported by the `OsFs`, but since that `Fs` can be used in others, they will also support it by proxy.

But not always, so hence this optional interface.

The interface is in this commit implemented for:

* BasePathFs
* OsFs
* CopyOnWriteFs
* ReadOnlyFs

Fixes #75

### DIFF
--- a/basepath.go
+++ b/basepath.go
@@ -9,6 +9,8 @@ import (
 	"time"
 )
 
+var _ Lstater = (*BasePathFs)(nil)
+
 // The BasePathFs restricts all operations to a given path within an Fs.
 // The given file name to the operations on this Fs will be prepended with
 // the base path before calling the base Fs.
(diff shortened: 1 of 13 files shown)


bep:
@spf13 as this is an API change it would be good if you could chime in.

* This is very well battle-tested in a branch in Hugo, so I'm fairly confident about the implementation.
* I'm also convinced that this is needed
* Symlink handling without something like this is not possible beyond the simple case

bep:
There is also https://github.com/spf13/afero/pull/75 -- which is a little bolder.

I don't think we can add this to the `Fs` interface; that would be a glass breaking change.

The `bool` return value is just used for testing, so we could make the signature into the more familiar `Lstat(name string) (os.FileInfo, error)` ... maybe.

bep:
This PR also contains a slightly unrelated new `UnionKey`-- see the commit message.

bep:
I'm talking to myself here, but just wanted to chime in to say that I'm in the "thinking box" about parts of this PR.

LandonTClipp:
Chiming in from the future.

I think this API change could have been done better. Because of the way this is implemented now, we have two modes in which a filesystem describes that it doesn't handle lstat-ing:
1. By simply not implementing `afero.Lstater`
2. By implementing `afero.Lstater` but returning `false` in the second return value.

This frustrates programming because you have to jump through not one, but now two hoops to determine if a filesystem actually supports symlinks (lstat'ing) "for real". Having that second bool parameter is trying to do the job that a type assertion would do. Why not just delegate it to a type assertion?

```go
type Lstater interface {
    Lstat(string) (os.FileInfo, error)
}

func (fs *MyCoolFs) Lstat(path string) (os.FileInfo, error) {
// do stuff
}
```

When callers want to know if Lstat is possible, they just do

```go
var stat os.FileInfo
lstater, ok := fs.(afero.Lstater)
if ok {
    stat = lstater.Lstat(path)
} else {
    stat = fs.Stat(path)
}
```

If callers don't care about whether or not lstat was actually called, they can make a function that wraps the above code block.

Unfortunately we can't change this now, but these are thoughts for the future.

PR #171, opened 2018-06-07
memmap: opening empty path should be an error

aviau:
@bep Can you confirm that I have updated the lstat tests correctly?

### DIFF
--- a/lstater_test.go
+++ b/lstater_test.go
@@ -73,7 +73,7 @@ func TestLstatIfPossible(t *testing.T) {
 	}
 
 	testLstat := func(l Lstater, pathFile, pathSymlink string) {
-		shouldLstat := pathSymlink != ""
+		shouldLstat := pathSymlink != FilePathSeparator
 		statRegular := checkLstat(l, pathFile, shouldLstat)
 		statSymlink := checkLstat(l, pathSymlink, shouldLstat)
 		if statRegular == nil || statSymlink == nil {
@@ -81,7 +81,7 @@ func TestLstatIfPossible(t *testing.T) {
 		}
 
 		symSym := statSymlink.Mode()&os.ModeSymlink == os.ModeSymlink
-		if symSym == (pathSymlink == "") {
+		if symSym == (pathSymlink == FilePathSeparator) {
 			t.Fatal("expected the FileInfo to describe the symlink")
 		}
 
@@ -95,8 +95,8 @@ func TestLstatIfPossible(t *testing.T) {
 	testLstat(overlayFs1, pathFile, pathSymlink)
 	testLstat(overlayFs2, pathFile, pathSymlink)
 	testLstat(basePathFs, "afero.txt", "symafero.txt")
-	testLstat(overlayFsMemOnly, pathFileMem, "")
-	testLstat(basePathFsMem, "aferom.txt", "")
+	testLstat(overlayFsMemOnly, pathFileMem, FilePathSeparator)
+	testLstat(basePathFsMem, "aferom.txt", FilePathSeparator)
 	testLstat(roFs, pathFile, pathSymlink)
-	testLstat(roFsMem, pathFileMem, "")
+	testLstat(roFsMem, pathFileMem, FilePathSeparator)
 }
--- a/memmap.go
+++ b/memmap.go
@@ -159,6 +159,11 @@ func (m *MemMapFs) MkdirAll(path string, perm os.FileMode) error {
 
 // Handle some relative paths
 func normalizePath(path string) string {
+	// Cleaning an empty path would result in "."
+	if path == "" {
+		return path
+	}
+
 	path = filepath.Clean(path)
 
 	switch path {
--- a/memmap_test.go
+++ b/memmap_test.go
@@ -449,3 +449,13 @@ func TestMemFsUnexpectedEOF(t *testing.T) {
 		t.Fatal("Expected ErrUnexpectedEOF")
 	}
 }
+
+func TestStatEmptyPath(t *testing.T) {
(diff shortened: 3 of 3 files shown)


bep:
I think you need to start with why this is needed? Looks like a breaking change.

aviau:
Because this is how the stdlib os package behaves.

bep:
>Because this is how Go's os package behaves. 

Go does not have a memory mapped filesystem in its Stdlib.

>This is only a breaking change for code that relied on a bug.

And that makes it all good? If this is a bug, I think it would be bad to punish the users who have followed the "current spec". 

I added a test in https://github.com/spf13/afero/pull/173 to illustrate my point. This is the `BaseFS´, which also behaves the same in this case. Whether this is wrong in that particular case, I'm not sure. You have a `BaseFS` that points at a relative location, so an empty string may make sense as the relative starting point (the root) of that file system.

But we can not change the above without breaking working applications. And I think we need a better reason than "this is how the OS filesystem behaves".

Also, you need to use the OS specific filepath separator in the tests.

aviau:
> And that makes it all good?

I think yes because otherwise afero couldn't be used to replace the os filesystem in unit tests without producing different results.

> so an empty string may make sense as the relative starting point (the root) of that file system.

Why? An empty string isn't accepted as the starting point of the os filesystem so why would it be different in Afero? The base, whether you are using BaseFS or not, should be "/".

> Also, you need to use the OS specific filepath separator in the tests.

done

aviau:
> I would say that writing unit tests against the OS root filesystem would be a rare thing

Indeed, so I use Afero instead.

I have a program that checks whether or not a file exists. The file name was empty but the program wouldn't notice because memMapFs.Stat("") does not return an error.

My tests run with MemMapFs but the program runs with OsFS.

The tests would pass but the program wouldn't work.

bep:
Are any of those issues fixed by this patch?

aviau:
> Are any of those issues fixed by this patch?

That isn't really what I meant. I think that the nature of these issues shows that afero's users expect that MemMapFs fail in cases where OsFs also fails. I was responding to your point that this would likely make people angry. If anything, I think that this will help people catch bugs or inconsistencies in their programs.

bep:
>I would say that writing unit tests against the OS root filesystem would be a rare thing. 

The above sentence does not make much sense, I meant: "unit tests against the root in the OS  filesystem".

aviau:
>  I meant: "unit tests against the root in the OS filesystem".

Does the example in my above comment answer your question?

I'll add code:

```golang

func DoSomethingWithFile(filename string, fs afero.FS) error {

    if err := fs.Stat(filename); err != nil {
        return err
    }

    // ...
}

```

This function would catch empty file names with OsFs but not with MemMapFs.

My unit tests use MemMapFs and didn't catch a bug where the caller would send an empty file name, but they should have!

bep:
>Does the example in my above comment answer your question?

Not really. Having client code interact directly with the root of the OS filesystem is most likely not what you would want. The beauty of Afero is no just the "unit test abstraction", there is a read only wrapper and a base filesystem, a composite etc. so you can restrict access to only the parts of the filesystem you want to expose.

So, for me, this issue falls into the big _maybe_ category as to "is this a bug" -- and in the big _probably too expensive_ in the "should we fix" category.

JoshuaKolden:
I think @aviau is correct that the expected behavior is consistency. In fact it is a clearly stated feature and use case in the readme.  Users of this package expect the statements in bold below to be accurate. They are not. Fixes are expected. Those relying on the current *broken* behavior and unable to update their usage should vendor the version they rely on. This is a very common workflow, and not an unreasonable expectation in the slightest. Otherwise all progress stops.

That being said there are quire a few bugs in the current code that need work.  I'd be happy to help if I had merge approval.

---
## Afero Features

* **A single consistent API for accessing a variety of filesystems**
* **Interoperation between a variety of file system types**
* **A set of interfaces to encourage and enforce interoperability between backends**
* An atomic cross platform memory backed file system
* Support for compositional (union) file systems by combining multiple file systems acting as one
* Specialized backends which modify existing filesystems (Read Only, Regexp filtered)
* A set of utility functions ported from io, ioutil & hugo to be afero aware


# Using Afero

Afero is easy to use and easier to adopt.

A few different ways you could use Afero:

* Use the interfaces alone to define you own file system.
* **Wrap for the OS packages.**
* Define different filesystems for different parts of your application.
* **Use Afero for mock filesystems while testing**

kargakis:
This looks like a good fix to me that will reduce confusion wrt expectations of how MkDir is supposed to work. It is a backwards-incompatible change but thankfully afero is using semantic versioning so the next tag to be created can be 2.0.0 to signal to users that there are backwards-incompatible changes. Unless @spf13 speaks up against a 2.x cut, I will do it within the next few days.

spf13:
My only advice on this is, if we are going to cut a 2.0 let's get all the
breaking changes we want into it. I'd also suggest announcing it and giving
a couple weeks for others to chime in / make suggestions for other breaking
changes.

kargakis:
Makes sense. Which is the best venue to announce so folks can chime in?

-------- Original Message --------

spf13:
Not sure. Everyone is distracted right now. Maybe give it more time given
that.

Probably twitter and the readme are good starts.

Maybe even reach out to some of the more prominent projects using Afero.
https://pkg.go.dev/github.com/spf13/afero?tab=importedby
Looks like a lot of the imports are through Viper, but there's a good
number of direct imports too.

the-native-developer (approved):

spf13:
Applied to master. Correct fix — empty path should error, not resolve to root.

PR #173, opened 2018-06-08
DO NOT MERGE: Add basefs test with empty root

bep:


aviau:
Why would an empty path be considered as the root? This is inconsistent with OsFs.

bep:
>Why would an empty path be considered as the root? This is inconsistent with OsFs

This test is just an example to make a point. The _why_ isn't important. One could possibly say that `OsFs` is the ugly duck here and should be aligned with the others, but that would probably not be OK.

aviau:
> One could possibly say that OsFs is the ugly duck here and should be aligned with the others

Right, I get your point.

I don't want to say that accepting an empty path would be a bad design. I understand that it could make sense.

However, in the case of afero, this should be considered as a bug because consistency with go's Os package adds much more value to the library than the convenience of accepting empty paths.

PR #177, opened 2018-09-28
Fix MemMapFs RealPath error when BasePath = ""

phyrwork:
bpath cleans to ".", but paths "./*" clean to "*".

This results in "." not found as prefix of "*" except
for hidden files and dirs.

Skip the prefix check if bpath == "." as this is
lexically equivalent to "", which would satisfy the
check.

### DIFF
--- a/basepath.go
+++ b/basepath.go
@@ -46,7 +46,7 @@ func (b *BasePathFs) RealPath(name string) (path string, err error) {
 
 	bpath := filepath.Clean(b.path)
 	path = filepath.Clean(filepath.Join(bpath, name))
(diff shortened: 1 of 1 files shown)


kargakis:
A unit test case that covers the change would be nice.

spf13:
Closing — superseded by #564 (the RealPath refactor, already merged). That PR changed from `strings.HasPrefix(path, bpath)` to `filepath.Rel(bpath, path)`, which handles the empty base-path case cleanly: `filepath.Rel(".", "foo")` returns `"foo"`, which doesn't start with `..`, so paths pass through correctly. No separate fix needed.

PR #181, opened 2018-11-19
Make constructors return structs instead of interfaces

### LINKED ISSUE #76: NewBasePathFs should return the concrete type

jribe:
This changes the return type of the various `NewXxxx` functions from `Fs` to the relevant type that's being returned.

To maintain the same level of compile-time checking, each type impacted now has a `var _ Fs = (*TheType)(nil)` no-op assignment to assert that the type implements `Fs`.

This change brings afero more in line with the ["accept interfaces, return structs" idiom](http://idiomaticgo.com/post/best-practice/accept-interfaces-return-structs/).

Importantly, this makes the constructor functions easier to discover in GoDoc since they will now be grouped with the structs that they build.

AFAIK, this should be completely backwards compatible despite changing the API because:

* The values are still assignable to `Fs`
* Interfaces are transparent to reflection

Edit: I suppose this would be a breaking change if anybody assigned the constructor to a value. Seems unlikely, but plausible. Would that require a major version bump?

Edit 2: This also breaks any code that performs a type assertion on the returned value (by making the assertion unnecessary).

Fixes https://github.com/spf13/afero/issues/76

### DIFF
--- a/basepath.go
+++ b/basepath.go
@@ -9,6 +9,7 @@ import (
 )
 
 var _ Lstater = (*BasePathFs)(nil)
+var _ Fs = (*BasePathFs)(nil)
 
 // The BasePathFs restricts all operations to a given path within an Fs.
 // The given file name to the operations on this Fs will be prepended with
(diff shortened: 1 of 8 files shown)


jeffwidman:
Be great to see this land

spf13:
This would be a significant breaking API change that removes the polymorphism at the heart of afero's design. The whole point of returning Fs and File interfaces is that callers can swap backends transparently. Returning concrete types would make all the compositional filesystems (BasePathFs, CopyOnWriteFs, etc.) incompatible with code using concrete types. Closing.

PR #196, opened 2019-02-16
Added ReverseHttpFs

dkumor:
There is already `HttpFs` which goes from an afero Fs to `http.FileSystem`. This adds the reverse: going from an `http.FileSystem` to an afero Fs.

The reason for this is that many tools expose `http.FileSystem` interfaces, since it is the standard in golang for serving files.

In particular, tools to embed files in golang executables usually return `http.FileSystem`s (see statik, packr, vfsgen, etc).

Allowing to convert these to Afero leads to some really useful things,
such as using CopyOnWriteFs to "overlay" a configuration folder/memory map over the file systems returned from such tools.

### DIFF
--- a/revHttpFs.go
+++ b/revHttpFs.go
@@ -1,0 +1,165 @@
+package afero
+
+import (
+	"io"
+	"net/http"
+	"os"
+	"syscall"
+	"time"
+)
+
+// ReverseHttpFile takes an http.File and matches it to File
+type ReverseHttpFile struct {
+	http.File
+}
+
+// Name returns the file's name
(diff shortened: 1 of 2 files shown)


julian7:
@dkumor why don't you release this patch as a separate lib? That would be short and sweet, as it can happily live outside of afero.

dkumor:
I guess that is the best option. I will do so once I am back from vacation.

dkumor:
I have released the patch as https://github.com/dkumor/revhttpfs

kargakis:
Since thoere is the external lib now there is no reason to keep this PR open.

PR #199, opened 2019-03-02
Fix UnionFile.Readdir return when c <= 0

### LINKED ISSUE #197: UnionFile.Readdir return io.EOF on empty dir when c=-1
A bug was recently introduced in https://github.com/spf13/afero/pull/195/commits/b158c092552c8569bb2dbc040553f286d3dc5b27.  `UnionFile.Readdir` doesn't properly handle the empty directory case where `f.off == len(f.files) == 0 && c == -1`.  `Readdir` should only return `io.EOF` if `c > 0`.

moorereason:
Update Readdir to match behavior of stdlib os package.

Fixes #197

### DIFF
--- a/composite_test.go
+++ b/composite_test.go
@@ -404,7 +404,7 @@ func TestCacheOnReadFsNotInLayer(t *testing.T) {
 }
 
 // #194
-func TestUniontFileReaddirEmpty(t *testing.T) {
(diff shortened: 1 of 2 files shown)


bep:
OK, sorry, I didn't see this ...

PR #200, opened 2019-03-05
Implement Symlinker

### LINKED ISSUE #160: Create symlink function
Does afero support or plan to support create symlink? This would be useful.
func Symlink(oldname, newname string) error

adracus:
This implements a `Symlinker` interface that supports the methods `EvalSymlinksIfPossible` and `SymlinkIfPossible`.

Fixes #160


kargakis:
Already implemented in https://github.com/spf13/afero/pull/228

PR #207, opened 2019-05-06
Added Support for openstack swift object storage

kolaente:
This pr adds support for openstack swift [object storage](https://docs.openstack.org/project-deploy-guide/openstack-ansible/newton/overview-storage-arch.html).

To run the tests, you'll need access to an openstack swift store, I've used [the one from ovh](https://www.ovh.ie/public-cloud/storage/object-storage/) since its pretty cheap (at least for testing). You'll then need to set the env variables `SWIFT_APIUSER`, `SWIFT_APIKEY`, `SWIFT_AUTHURL` and `SWIFT_DOMAIN`. Also someone needs to do this in travis if we want to run these tests in the ci.


kargakis:
Needs a rebase

kolaente:
@kargakis Done!

kolaente:
I force-pushed the changes again since I kind messed up the rebase and accidentally included a lot of unrelated changes in the branch. Now it's all nice and clean.

kargakis:
Can you document how to configure a Swift backend in https://github.com/spf13/afero#available-backends? Also the unit tests seem to fail in CI, please have a look.

kolaente:
The test fails in ci because it needs a swift store to run the test and no access keys were provided. Someone with access to the ci needs to set the env variables `SWIFT_APIUSER`, `SWIFT_APIKEY`, `SWIFT_AUTHURL` and `SWIFT_DOMAIN`.

I'll add docs about how to use the backend.

kargakis:
@spf13 please have a look and let us know how do you want to proceed with unit testing the Swift backend.

kolaente:
@kargakis Added the documentation.

IMHO, "classic" unit testing this does not really make sense since most of the package is glue code between afero and the swift package. I think this kind of integration test with talking to a "real" storage server makes the most sense. Maybe we could put one (through docker?) in the ci though to reduce dependency on external services. Not sure how much of a hassle that would be.

kargakis:
> IMHO, "classic" unit testing this does not really make sense since most of the package is glue code between afero and the swift package. I think this kind of integration test with talking to a "real" storage server makes the most sense. Maybe we could put one (through docker?) in the ci though to reduce dependency on external services. Not sure how much of a hassle that would be.

This is nitpicking but you are right, my bad on using the word "unit". In any case, my question to @spf13 still stands, I am not sure how easy or whether it is possible at all to run our own services in Travis and/or AppVeyor.

kargakis:
As with the GCS implementation, we can go ahead and merge this without "real" storage tests, just need to note in README that there is **experimental** support.

https://github.com/spf13/afero/pull/331

kargakis:
@kolaente can you rebase on top of latest master?

kolaente:
@kargakis wow I completely forgot I had this PR open :upside_down_face: 

Just rebased, what should we do with the tests? Since they can't really run in CI as of now.

kargakis:
Does the Swift library you are using have a mock client?

kolaente:
The swift library has a [fake swift server](https://github.com/ncw/swift/blob/master/swift_test.go#L6-L7) which stores everything in memory. As far as I understand it, that's kind of a mock client but more on the integration side. Not sure if we can use this here.

spf13:
Thank you for building this. Afero's design philosophy is to be a thin, composable abstraction layer - the core package ships a small set of general-purpose filesystem types and a handful of well-maintained backend drivers. We've decided not to expand that set with additional external-service backends inside this repo, because each one adds a permanent maintenance burden, external service dependency, and CI complexity.

The right home for this is a standalone module published under your own namespace (e.g. github.com/mkvolkov/afero-rclone for Rclone, github.com/kolaente/afero-swift for Swift, etc.). That gives you full control over releases and dependencies, and makes it immediately usable without pulling additional dependencies into afero itself.

We'd be happy to list community backends in the README. If you publish it as a standalone module, open a small PR adding a line to the README and we'll merge it promptly.

PR #211, opened 2019-07-11
Adjust MemFs.List() to return the data

grongor:
I found the List method quite useless before. I was also unable to properly test my code without this method - I would have to iterate through entire filesystem to ensure that my code didn't write to some unexpected place.

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -14,13 +14,10 @@
 package afero
 
 import (
-	"fmt"
 	"io"
-
 	"log"
 	"os"
 	"path/filepath"
-
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


grongor:
So...3 years and the `MemMapFs` is still unusable for many tests because I can't query it in any way. Don't you think it's time to fix this? Printing something in the library wasn't ever a clever idea. If the "breaking change" is the issue then let's just choose another name for the method.

wuxia101:
This adjustment is absolutely necessary 👍 👍 👍

wuxia101 (commented):
This is exactly what I want

wuxia101:
It would be better to use this change. https://github.com/spf13/afero/pull/589

spf13:
Closing. The current `List()` method that prints to stdout is clearly bad API design, and changing its return type is the right direction. But this PR is 7 years old, needs a full rebase given all the memmap.go changes since then, and conflicts with #589. We'll fix `List()` properly ourselves in a cleanup pass. In the meantime, `Walk()` is the right tool for iterating all files programmatically.

Issue #217, opened 2019-08-30
File extension for afero.TempFile


aswinkarthik:
Currently, we are unable to set file extension for a temp file that is created using `afero.TempFile`. In `ioutil.TempFile` we are able to do so by specifying the pattern like `prefix-*.yml`, the temp file will be created with a random string at the last `*`. Refer [here](https://golang.org/pkg/io/ioutil/#TempFile)

Can `afero.TempFile` also do a similar replacement?  If this change can be accepted, will love to raise a PR.

nithyanatarajan:
@spf13 We are also in need of the feature. I have raised a PR for the same. Kindly have a look at it.

kargakis:
Fixed by https://github.com/spf13/afero/pull/229

PR #221, opened 2019-10-05
Modify afero.TempFile() to support pattern instead of prefix

nithyanatarajan:
Context:

- initially afero.TempFile() takes prefix as last argument and adds a random string to the end. This posed difficulty when trying to add a file extension with randomised filename.
- to avoid this, going forward afero.Tempfile() takes a pattern and replace the last * with the random string. This behavior is same as ioutil.Tempfile()

This PR closes the issue https://github.com/spf13/afero/issues/217


kargakis:
oops, sorry for missing this PR, already fixed in https://github.com/spf13/afero/pull/229 but thanks for the contribution!

PR #223, opened 2019-11-03
Remove stale release notes from README

whereswaldon:
New users coming to the project could incorrectly assume that the project is pre-1.0 release and that it hasn't had a release in years if they assumed that the release notes in the README were up-to-date. I say this because I assumed that on my first read.


whereswaldon:
Finally remembered to circle back on that CLA.

PR #226, opened 2019-11-11
Proper root init for MemMapFs in windows

### LINKED ISSUE #225: MemMapFs open windows absolute path does not exist
Doing `fs.Open("c:\\")` or `afero.Walk` from root on a MemMapFs in a windows system returns the error `open c:\: file does not exist`. During debugging, I found that the `getData()` function in MemMapFs does not support windows root as shown by the TODO in it (https://github.com/spf13/afero/blob/master/memmap.go#L38). 

I would like to fix it in my fork, if it is okay.

nofun97:
Fixes #225 

### Problem:
MemMapFs is set to initialize with a map data structure and a root path was added to it. The path that was added uses the `os.PathSeparator` as the root path. While UNIX has no problem with this, Windows does as the volume name is not present in `os.PathSeparator`

### Changes:
* MemMapFs `getData()` init is now changed to add a root absolute instead of `FilePathSeparator`
* `normalizePath(path string) string` is changed to return a proper root absolute if the cleaned path results in a root path (`/` or `\\`) and an error
* BasePathFs removes volume name from path when `Name() string` is called to hide real path
* Several tests are changed to implement the changes

### DIFF
--- a/composite_test.go
+++ b/composite_test.go
@@ -479,7 +479,7 @@ func TestUnionFileReaddirAskForTooMany(t *testing.T) {
 	overlay := &MemMapFs{}
 
 	for i := 0; i < 5; i++ {
-		WriteFile(base, fmt.Sprintf("file%d.txt", i), []byte("afero"), 0777)
+		WriteFile(base, fmt.Sprintf("/file%d.txt", i), []byte("afero"), 0777)
 	}
 
 	ufs := &CopyOnWriteFs{base: base, layer: overlay}
--- a/composite_test.go
+++ b/composite_test.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"path/filepath"
+	// "path/filepath"
 	"os"
 	"testing"
 	"time"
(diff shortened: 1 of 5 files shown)


composite_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	// "path/filepath"
```
kargakis: remove

composite_test.go
```diff
@@ -478,11 +479,11 @@ func TestUnionFileReaddirAskForTooMany(t *testing.T) {
 	overlay := &MemMapFs{}
 
 	for i := 0; i < 5; i++ {
-		WriteFile(base, fmt.Sprintf("file%d.txt", i), []byte("afero"), 0777)
+		WriteFile(base, fmt.Sprintf("/file%d.txt", i), []byte("afero"), 0777)
 	}
 
 	ufs := &CopyOnWriteFs{base: base, layer: overlay}
-
+
```
kargakis: Can you please gofmt your code?

nofun97:
rebased and added requested changes

composite_test.go
nofun97: removed

composite_test.go
nofun97: right, must have misconfigured my file watcher, it's removed.

memmap.go
```diff
@@ -38,17 +38,26 @@ func NewMemMapFs() Fs {
...
 	m.init.Do(func() {
 		m.data = make(map[string]*mem.FileData)
-		// Root should always exist, right?
-		// TODO: what about windows?
-		m.data[FilePathSeparator] = mem.CreateDir(FilePathSeparator)
+		// Root should always exist
+		absolutePath, err := filepath.Abs(FilePathSeparator)
+		if err != nil {
+			log.Fatal(err)
+			panic(err)
```
kargakis: Is panic needed if you log.Fatal above?

kargakis:
This looks like a breaking change so need other folks to chime in and comment on whether it makes sense to have.

kargakis:
See also https://github.com/spf13/afero/pull/171

spf13:
Closing this 6-year-old PR. The core Windows path concerns have been partially addressed by other merges since then (#564 for BasePathFs, etc.). A few issues with this approach that make it hard to take as-is: (1) the double-fault pattern of log.Fatal + panic, (2) changing normalizePath's signature to return (string, error) is a large internal refactor, and (3) storing the root as filepath.Abs('/') on Windows would pin the in-memory fs to a specific drive letter, which is surprising. If Windows MemMapFs root handling is still broken, a fresh focused PR would be much easier to review.

Issue #141, opened 2017-10-20
MemMapFS doesn't properly rename directories


kklin:
When I rename a directory with MemMapFS, the files within it don't get moved. I _think_ the files within should also move to the new directory name (as in the behavior of `mv <old_dir> <new_dir>`), but maybe I'm misunderstanding how Rename works.

I've written a unit test that shows the error:
```
// TestRenameDirectory creates a directory with a file, and attempts to rename
// the directory. The child file should also be within the renamed directory.
func TestRenameDirectory(t *testing.T) {
        fs := NewMemMapFs()

        srcDir := "/src-dir"
        dstDir := "/dst-dir"
        filename := "file"
        fileInSrcDir := filepath.Join(srcDir, filename)
        fileInDstDir := filepath.Join(dstDir, filename)

        if err := fs.Mkdir(srcDir, 755); err != nil {
                t.Fatalf("Mkdir failed unexpectedly: %s", err)
        }

        if _, err := fs.Create(fileInSrcDir); err != nil {
                t.Fatalf("Create file failed unexpectedly: %s", err)
        }

        if err := fs.Rename(srcDir, dstDir); err != nil {
                t.Fatalf("Rename failed unexpectedly: %s", err)
        }

        if _, err := fs.Stat(fileInSrcDir); err == nil {
                t.Error("The file should not be in the old directory")
        }

        if _, err := fs.Stat(fileInDstDir); err != nil {
                t.Errorf("The file should be in the new directory: %s", err)
        }
}
```

```
--- FAIL: TestRenameDirectory (0.00s)
        memmap_test.go:371: The file should not be in the old directory
        memmap_test.go:375: The file should be in the new directory: open /dst-dir/file: file does not exist
```

I could help out fixing this, but I'm not sure how to approach it. I took a quick glance at the way the filesystem is represented within MemMapFS, and didn't completely understand it.

Thanks!

mtaufen:
I also noticed this recently while unit testing something that moves directories with MemMapFS; it would work if I tested with the OS default filesystem, but not with MemMapFS.

bclermont:
Confirm

```
func TestAferoRename(t *testing.T) {
	fs := afero.NewMemMapFs()
	require.NoError(t, fs.MkdirAll("/d1/d2/d3", 755))
	f, err := fs.OpenFile("/d1/d2/d3/f1", os.O_CREATE|os.O_WRONLY, 0644)
	require.NoError(t, err)
[...]

PR #239, opened 2020-04-11
Fix renaming a dir with sub-directories

nono:
I think it should fix issue #141

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -319,6 +319,18 @@ func (m *MemMapFs) Rename(oldname, newname string) error {
 	} else {
 		return &os.PathError{Op: "rename", Path: oldname, Err: ErrFileNotFound}
 	}
+
+	for p, fileData := range m.getData() {
+		if strings.HasPrefix(p, oldname+FilePathSeparator) {
+			m.mu.RUnlock()
+			m.mu.Lock()
+			delete(m.getData(), p)
+			p := strings.Replace(p, oldname, newname, 1)
+			m.getData()[p] = fileData
+			m.mu.Unlock()
+			m.mu.RLock()
+		}
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


memmap_test.go
```diff
@@ -449,3 +449,47 @@ func TestMemFsUnexpectedEOF(t *testing.T) {
...
+	const dstPath = "/dst"
+	const subDir = "dir"
+
+	fs := NewMemMapFs()
+
+	err := fs.MkdirAll(srcPath+FilePathSeparator+subDir, 0777)
+	if err != nil {
+		t.Errorf("MkDirAll failed: %s", err)
+		return
+	}
```
kargakis, line 464: Can you also add a file in the subdirectory and not have only directories in the test?

memmap_test.go
```diff
@@ -449,3 +449,47 @@ func TestMemFsUnexpectedEOF(t *testing.T) {
...
+
```
kargakis, line 465: Also please make sure to stat the file here before renaming

memmap.go
```diff
@@ -270,7 +270,7 @@ func (m *MemMapFs) RemoveAll(path string) error {
 	defer m.mu.RUnlock()
 
 	for p := range m.getData() {
-		if strings.HasPrefix(p, path) {
+		if strings.HasPrefix(p, path+FilePathSeparator) {
```
kargakis, line 273: Why is this change needed?

nono:
@bep hi, this is an old pull request but I think it would still be a good fix for afero. I have rebased it and listened to the comments made by @0xmichalis. If you have time to look at it, it would be appreciated.

nono:
Thanks :heart:

hanagantig:
It doesn't fix the issue https://github.com/spf13/afero/issues/141
I have described the problem here https://github.com/spf13/afero/issues/141#issuecomment-1530047212

You can check my PR which fix it https://github.com/spf13/afero/pull/364

PR #247, opened 2020-06-26
Fix panic when not filling up zipfs's read buffer

JohnStarich:
Thank you for this fantastic library. This has made my life _so much easier_. I tried `go-git/go-billy` for while, but it lacked feature-completeness on the in-memory FS. Afero did that better, and more 😄 

Here's a bug fix for a panic I ran into when reading small files into large byte buffers. It appears the offset increment was a tad off.


JohnStarich:
Thanks @kargakis!

PR #249, opened 2020-07-12
Prevent MemMapFs.Chmod from changing all mode bits

JohnStarich:
I ran into an issue where my program called `fs.Chmod` and it changed from a directory to a regular file. This one had me stumped for weeks 😅 

After checking [`os.Chmod`'s docs](https://golang.org/pkg/os/#Chmod), I bit-masked the part Chmod is permitted to set, then OR'd it with the file's other previous mode bits. I also added a test to check the file type after setting all permission bits to 0.

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -141,9 +141,7 @@ func (m *MemMapFs) Mkdir(name string, perm os.FileMode) error {
 	m.registerWithParent(item)
 	m.mu.Unlock()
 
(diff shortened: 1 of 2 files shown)


kargakis:
Good catch @JohnStarich, I have released [v1.3.2](https://github.com/spf13/afero/releases/tag/v1.3.2)

PR #250, opened 2020-07-15
Improve handling of mode bits

anishathalye:
- "/" should have mode `os.ModeDir|0755`, not `0000`. Among other things, this had resulted in `mode.IsDir()` returning false for root prior to this patch.
- `Mkdir`, `MkdirAll`, and `OpenFile` shouldn't be allowed to set permissions that are otherwise illegal through `Chmod`. This mirrors what Go's `os` package does: it calls `syscallMode(mode)`, which effectively clears out the same bits that are disallowed by `Chmod`.
- `MkdirAll` should use the given permissions for all intermediate directories that are created, not just for the final directory. Prior to this patch, intermediate directories were created with mode bits `0000`. Besides the permission bits being wrong, `mode.IsDir()` would return false for these directories prior to this patch.

A potentially amusing side note: I was working around these issues in my code by abusing the buggy `fs.Chmod()`, but that was fixed in #249.

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -25,6 +25,8 @@ import (
 	"github.com/spf13/afero/mem"
 )
 
+const chmodBits = os.ModePerm | os.ModeSetuid | os.ModeSetgid | os.ModeSticky // Only a subset of bits are allowed to be changed. Documented under os.Chmod()
+
 type MemMapFs struct {
 	mu   sync.RWMutex
 	data map[string]*mem.FileData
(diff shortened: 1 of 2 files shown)


JohnStarich:
Nice, this is a great follow-up to the illegal chmod fix. (I'd approve if I could, so poking @kargakis)

Interesting tidbit, Go itself doesn't support some of those bits for `os.Mkdir` (for no apparent reason): https://github.com/golang/go/issues/25539

kargakis:
Released in [v1.3.3](https://github.com/spf13/afero/releases/tag/v1.3.3)

PR #262, opened 2020-08-20
Fix MemMapFs.Readdir to use files buffer

### LINKED ISSUE #261: Directory buffering in MemMapFs.Readdir
`os.File.Readdir` under the hood uses buffering of directory contents, which gives it ability to traverse directory and remove files at the same time. Currently, `MemMapFs.Readdir` retrieves files from `f.fileData.memDir.Files()` on each call, which means that if one tries to remove a file between `Readdir` calls, `f.readDirCount` won't be relevant and in some cases slice bounds might be violated.

std0:
To match behavior of `os.File.Readdir`, `MemMapFs.Readdir` should use files buffer instead of retrieving currently stored files on each call.

Fixes #261

### DIFF
--- a/mem/file.go
+++ b/mem/file.go
@@ -31,6 +31,7 @@ type File struct {
 	// atomic requires 64-bit alignment for struct field access
 	at           int64
 	readDirCount int64
+	dirBuf       []*FileData
 	closed       bool
 	readOnly     bool
 	fileData     *FileData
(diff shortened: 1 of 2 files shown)


kargakis:
Needs a rebase to fix conflicts

std0:
@kargakis, done.

spf13:
Applied to master. Thanks for the careful fix.

PR #264, opened 2020-09-07
Make MemMapFs implement Lstater

LandonTClipp:
MemMapFs does not currently implement Lstater which can make it a bit difficult when a program is wanting to call Lstat on an unknown filesystem, so I am implementing the interface. The real call is to `Stat()`, and a `false` is simply returned to indicate that an Lstat was not called.

This change was prompted in my work of https://github.com/chigopher/pathlib, specifically: https://github.com/chigopher/pathlib/pull/16. The issue arose when testing a Walk function that calls LstatIfPossible, where the test uses a MemMapFs.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,7 @@ module github.com/spf13/afero
 
 require (
 	github.com/pkg/sftp v1.10.1
+	github.com/stretchr/testify v1.4.0
 	golang.org/x/crypto v0.0.0-20190820162420-60c769a6c586
 	golang.org/x/text v0.3.3
 )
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -2,6 +2,7 @@ module github.com/spf13/afero
 
 require (
 	github.com/pkg/sftp v1.10.1
+	github.com/stretchr/testify v1.4.0
```
kargakis, line 5: I would prefer this to either be used across all unit tests or not at all. Given the scope of your PR, I would remove it for now (feel free to open an issue to track updating all unit tests to use testify).

LandonTClipp: No worries, removed the use of testify and just relying on `t.Fatal(...)` for now.

kargakis (approved):

PR #265, opened 2020-09-12
Add tarfs implementation

agimenez:
Hi,

Heavily inspired by the already existing zipfs implementation, I have implemented the "tar" version. It is a read-only, in-memory representation of the contents of a tarball.

Thanks!


agimenez:
For some reason this is failing in AppVeyor, but I can't reproduce the panic in my development environment. Is there something I can do to fix it?

tarfs/tarfs_test.go
```diff
@@ -0,0 +1,402 @@
...
+	{"/", []string{"sub", "testDir1", "testFile"}},
+	{"/sub", []string{"testDir2"}},
+	{"/sub/testDir2", []string{"testFile"}},
+	{"/testDir1", []string{"testFile"}},
+}
+
+var afs *afero.Afero
+
+func TestMain(m *testing.M) {
+	tf, err := os.Open("testdata/t.tar")
```
kargakis, line 49: You probably have to fix this path to work for Windows in order to make appveyor pass.

agimenez: Hi, thanks! I think that this is OK, the panic is how I handle the "pseudo-root" entry internally, as I use a literal slash, and to be honest I didn't think on Windows systems. I set up an AppVeyor account and will debug the issue.

agimenez:
Hi, I'm closing this PR until I have the tests passing on AppVeyor, will reopen it when it's fixed

PR #268, opened 2020-10-05
Do not panic on passing empty tar reader to tarfs

hasheddan:
Corrects a bug where passing an empty tar reader to tarfs would cause
adding the pseudoroot to panic with assignment to entry in nil map.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

cc @agimenez because it looked like you implemented tarfs (thank you for your work!)

@kargakis would it be possible to get a v1.4.1 published after merge?

### DIFF
--- a/tarfs/fs.go
+++ b/tarfs/fs.go
@@ -63,6 +63,9 @@ func New(t *tar.Reader) *Fs {
 
 	}
 
+	if fs.files[afero.FilePathSeparator] == nil {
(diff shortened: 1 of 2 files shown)


agimenez:
Good catch! LGTM, thanks!

kargakis (approved):

kargakis:
[v1.4.1](https://github.com/spf13/afero/releases/tag/v1.4.1) is published

hasheddan:
Thanks @agimenez and @kargakis!

PR #291, opened 2021-03-13
Add adapter to allow afero.Fs usage as io/fs.FS

### LINKED ISSUE #275: go 1.16 compatibility
It seems that future releases of go will bring us new io/fs package with some interfaces for file system (https://tip.golang.org/pkg/io/fs). 
Is it possible to make afero compatible with io/fs?

xakep666:
It will not affect users of go older than 1.16 because files with `io/fs` import marked with build constraints

Will close #275


kargakis:
Released as part of https://github.com/spf13/afero/releases/tag/v1.6.0

jeffwidman:
> It will not affect users of go older than 1.16 because files with io/fs import marked with build constraints

Unfortunately, that's not entirely true... see https://github.com/golang/go/issues/44557

I'm not saying you should revert, as end users can simply wait to upgrade til the [backport to go 1.15](https://github.com/golang/go/issues/44792) lands. 

But did want to make folks aware.

PR #302, opened 2021-05-09
Fix long relative path handling on windows

sandreas:
On windows long relative paths led to an error (see https://github.com/golang/go/issues/21782). Here is a code example that fails (only on windows, on play it works): https://play.golang.org/p/wMS5gti4SD

To reproduce it you may use this on windows 10:

```
mkdir C:\inetpub\wwwroot\something_4.0\node_modules\babel-preset-es2015\node_modules\babel-plugin-transform-es2015-block-scoping\node_modules\babel-traverse\node_modules\babel-code-frame\node_modules\chalk\node_modules\strip-ansi\node_modules\ansi-regex\
echo "" > C:\inetpub\wwwroot\something_4.0\node_modules\babel-preset-es2015\node_modules\babel-plugin-transform-es2015-block-scoping\node_modules\babel-traverse\node_modules\babel-code-frame\node_modules\chalk\node_modules\strip-ansi\node_modules\ansi-regex\test.txt
```

And then use the `afero.Walk` on `C:\inetpub\wwwroot\something_4.0\` with a relative path traversal (e.g. `.` or `node_modules`).

Other projects solve this problem in a similar way:
- `rclone`: https://github.com/rclone/rclone/blob/9e2fbe0f1a9314e1330b4adc9d69de137be12ee6/lib/file/unc_windows.go#L16
- `restic`: https://github.com/restic/restic/blob/74c0607c9222edec3b0c140bb6fee962d6d2e82d/internal/fs/file_windows.go#L12

### DIFF
--- a/os.go
+++ b/os.go
@@ -34,7 +34,7 @@ func NewOsFs() Fs {
 func (OsFs) Name() string { return "OsFs" }
 
 func (OsFs) Create(name string) (File, error) {
-	f, e := os.Create(name)
+	f, e := os.Create(normalizeLongPath(name))
 	if f == nil {
 		// while this looks strange, we need to return a bare nil (of type nil) not
 		// a nil value of type *os.File or nil won't be nil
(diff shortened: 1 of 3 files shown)


sandreas:
Is there any plan to merge this or give feedback? Otherwise I can close this...

sandreas:
Closed due to no feedback within > 18 month.

PR #303, opened 2021-05-10
Make uid and gid readable.

iineva:


spf13:
Closing — superseded by #349, which we're applying to master. Both PRs implement `Sys()` returning `*syscall.Stat_t` with uid/gid on Unix. #349 is cleaner: it uses modern `//go:build` syntax (vs old-style `// +build`), handles the wasm platform explicitly, and has been more recently maintained. Thanks for raising this — the fix is landing.

PR #305, opened 2021-05-10
Issue 304 - introducing early check for pattern as done in filepath.Glob

jbaudens:


jbaudens:
It looks like my tests fail here because the CI is using go 1.13 when my local env is using 1.16? Looks like there has been changes in the filepath package in go after 1.13.

PR #309, opened 2021-06-19
Add new filtering fs: FilePredicateFs

satotake:
While golang's built-in `regexp` package has the advantage, [constant time guarantees](https://golang.org/pkg/regexp/),
it (RE2) has some limitations to express complex rules compared to other languages (PCRE2).
(For example, it does not have backtracking.)
Here, I would like to introduce a new filtering backend, `PredicateFs`, to make afero's filtering more flexible and composable.

`PredicateFs` is a filtered view on predicates which takes file path as an argument.
Any file will be treated as non-existing when the predicate returns false.
Unlike `RegexpFs`, the fs targets not file name but file path.
Like, `RegexpFs`, files will not be created when the predicate returns false and directories are always not filtered.

### DIFF
--- a/predicatefs.go
+++ b/predicatefs.go
@@ -1,6 +1,7 @@
 package afero
 
 import (
+	"io/fs"
 	"os"
 	"path/filepath"
 	"syscall"
@@ -13,6 +14,10 @@ type FilePredicateFs struct {
 	pred   func(string) bool
 	source Fs
 }
+
+var (
+	_ fs.ReadDirFile = (*PredicateFile)(nil)
+)
 
 func NewFilePredicateFs(source Fs, pred func(string) bool) Fs {
 	return &FilePredicateFs{source: source, pred: pred}
@@ -189,6 +194,25 @@ func (f *PredicateFile) Readdir(c int) (fi []os.FileInfo, err error) {
 		}
 	}
 	return fi, nil
+}
+
+func (f *PredicateFile) ReadDir(n int) (filtered []fs.DirEntry, err error) {
+	var entreis []fs.DirEntry
+	if rdf, ok := f.f.(fs.ReadDirFile); ok {
+		entreis, err = rdf.ReadDir(n)
+	} else {
+		entreis, err = readDirFile{f.f}.ReadDir(n)
+	}
+	if err != nil {
+		return nil, err
+	}
+	for _, e := range entreis {
+		if e.IsDir() || f.pred(filepath.Join(f.f.Name(), e.Name())) {
+			filtered = append(filtered, e)
+		}
+	}
+	return filtered, nil
+
 }
 
 func (f *PredicateFile) Readdirnames(c int) (n []string, err error) {
--- a/predicatefs.go
+++ b/predicatefs.go
@@ -7,15 +7,15 @@ import (
 	"time"
 )
 
-// PredicateFs filters files (not directories) by predicate,
+// FilePredicateFs filters files (not directories) by predicate,
 // which takes file path as an arg.
-type PredicateFs struct {
+type FilePredicateFs struct {
 	pred   func(string) bool
 	source Fs
 }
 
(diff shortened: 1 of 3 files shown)


jxsl13:
This one is pretty nice as it may filter according to any kind of path related rule, not just a regular expression.

satotake:
> You removed a #from the readme of HttpFs

This is not related with this request directly but I think HttpFs is not filtering fs.

> I wonder what the thoughts behind that reasoning were?
> Same as RegexpFs?

Good point.
Though I am not sure of the rationales of RegexpFs, probably they are the same. 

Consider a situation where we

- want to include only files with `.go` extension
- use `filepath.Walk`, which calls `Open` internally

(yes, this will be minor)

If we apply the function with both files and dirs, `strings.HasSuffix(s, ".go")` will be wrong for files in sub dir.
`dir/example.go` is skipped unexpectedly because `strings.HasSuffix("dir", ".go")` is false.
This is easy to go wrong in my opinion. 

Instead of it, as you describe, it will be clear to combine `PredicateFs` with `BasepathFs` etc.
Or, we can handle directories as I wrote the test though it is a little verbose (`.hidden`).

jxsl13:
In case that you provide the underlying `Fs` as parameter to the predicate function, you might be able to check whether the file path is actually a file or a directory with a `Stat` call and then go for the file's suffix.

^
I think that this would make things way more complicated, for sure and potentially more error prone.

Might as well rename `PredicateFs` to `FilePredicateFs` in order to make it clearer that this abstraction is only targeting files.

satotake:
@jxsl13 thanks for reviewing.
Replaced `PredicateFs` with `FeilePredicateFs`

bep:
Sorry for being very late to the game, this looks very useful. I have read the discussion above, but I suspect that we're missing out on some potential when we apply this to files (and not dirs) only. If I walk a directory I certainly want to be able to skip a dir before reading it.

How about making the predicate:

```go
func(isDir bool, path string) bool
```

satotake:
@bep thank you.
I referred to `RegexpFs` and it skips all directories internally.
So does this fs.

predicatefs.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func (f *PredicateFile) Readdir(c int) (fi []os.FileInfo, err error) {
+	var pfi []os.FileInfo
+	pfi, err = f.f.Readdir(c)
+	if err != nil {
+		return nil, err
+	}
+	for _, i := range pfi {
+		if i.IsDir() || f.pred(filepath.Join(f.f.Name(), i.Name())) {
```
bep, line 187: @satotake I'm not sure what you mean by "skip directories"; to me it looks like every directory is included.

I would have found it much more useful if the above was replaced with:

```go
if f.pred(i.IsDir(), filepath.Join(f.f.Name(), i.Name())) {
```

jxsl13: One might as well just pass the file info object instead of a bool which can then be manually checked whether the path is a directory or a file or a symlink, etc.

bep: That would break if you want to use this with `fs.ReadDirFile` (see my recent commit to the main branch).

satotake: @bep I got your point.
`isDir` will be nice to ignore some dirs without scanning their children, for instance.
Fixed.

spf13:
Closing. The `validate()` function calls `IsDir()` which stats the file to determine its type — but for `Create()` on a non-existent path, that stat fails, making the predicate unworkable for the creation case. That's a fundamental correctness issue.

The predicate function signature `func(bool, string) bool` (taking an isDir flag computed by statting the file) is also awkward: the predicate can only be evaluated after the file exists, which inverts the intended semantics for creation.

`RegexpFs` covers the common filtering use case. For full-path predicate filtering, this can be composed externally without a new FS type in core. Thanks for the contribution — the idea of path-based vs name-based filtering is sound, but the implementation needs a different approach.

PR #310, opened 2021-06-19
regexpfs: always use file name instead of path

satotake:
On `RegexpFs`, some operations apply given regexps with not file **name** but file **path**.
Fix this.

README
> A filtered view on file *names*, any file NOT matching the passed regexp will be treated as non-existing.

### DIFF
--- a/regexpfs.go
+++ b/regexpfs.go
@@ -2,6 +2,7 @@ package afero
 
 import (
 	"os"
+	"path/filepath"
 	"regexp"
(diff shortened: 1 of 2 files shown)


jxsl13:
according to the description of RegexpFs this fix makes sense.

(imo RegexpFs should be more powerful to also match full file paths but that's another discussion)

spf13:
Applied to master. This was clearly the right behavior per the docs.

PR #313, opened 2021-07-09
Readme: fix quotes of step 3

RaphaelPour:
Single quotes are reserved for runes and therefore not allowed for strings in Go: `more than one character in rune literal`


README.md
```diff
@@ -127,7 +127,7 @@ WriteString(s string) : ret int, err error
 In some applications it may make sense to define a new package that
 simply exports the file system variable for easy access from anywhere.
 
-## Using Afero's utility functions
+## Using Aferos utility functions
```
satotake, line 130: This quote should be allowed though I am not maintainer

kargakis: Agreed

PR #320, opened 2021-09-02
(GH-141) Fix renaming a dir with sub-directories

da-ar:
This is an update to https://github.com/spf13/afero/pull/239 to address the open comments, and rebased off master

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -279,7 +279,7 @@ func (m *MemMapFs) RemoveAll(path string) error {
 	defer m.mu.RUnlock()
 
 	for p := range m.getData() {
-		if strings.HasPrefix(p, path+FilePathSeparator) {
+		if strings.HasPrefix(p, path) {
 			m.mu.RUnlock()
 			m.mu.Lock()
 			delete(m.getData(), p)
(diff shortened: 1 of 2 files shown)


jxsl13:
pls merge.

jxsl13:
what happens if you have got 
a sub directory called 

```
/subdir1/whatever/
/subdir10/whatever1/
```
and you try to rename `subdir1`?

spf13:
The fix is in the right direction but has two bugs that need addressing before this can merge:

1. **Data race in the range loop.** You're doing a lock upgrade (`RUnlock → Lock → mutate → Unlock → RLock`) inside a `range` over the same map you're mutating. This is a data race: the range iterator's internal state is invalidated by the concurrent mutation. Collect the keys to rename first, then mutate in a second pass:
   ```go
   // collect under read lock
   m.mu.RLock()
   var toRename []string
   for p := range m.getData() {
       if strings.HasPrefix(p, oldname+"/") || p == oldname {
           toRename = append(toRename, p)
       }
   }
   m.mu.RUnlock()
   // mutate under write lock
   m.mu.Lock()
   for _, p := range toRename {
       newP := newname + p[len(oldname):]
       m.getData()[newP] = m.getData()[p]
       delete(m.getData(), p)
   }
   m.mu.Unlock()
   ```

2. **Prefix collision.** `strings.HasPrefix(p, oldname)` matches `/subdir10` when renaming `/subdir1`. The prefix check must be `strings.HasPrefix(p, oldname+"/")` (or `normalizePath(oldname)+FilePathSeparator`).

Fix both and this is mergeable.

spf13:
Thanks for this — the bug you identified was real. However, this fix is now already in the codebase via renameDescendants(), which was added in a subsequent commit and handles this case correctly (and avoids the path-prefix collision that would affect your HasPrefix check, e.g. renaming /src incorrectly matching /src-other). Closing as already fixed.

PR #323, opened 2021-09-12
os: Include Go 1.16 and Go 1.17 os package functions

Foxboron:
After the deprecations of ioutil most of the new functions has been
migrated to the os package.

This patch includes all the new functions.

Signed-off-by: Morten Linderud <morten@linderud.pw>

### DIFF
--- a/os.go
+++ b/os.go
@@ -15,6 +15,7 @@
 package afero
 
 import (
+	"io/fs"
 	"os"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


os.go
```diff
@@ -111,3 +120,15 @@ func (OsFs) SymlinkIfPossible(oldname, newname string) error {
 func (OsFs) ReadlinkIfPossible(name string) (string, error) {
 	return os.Readlink(name)
 }
+
+func (OsFs) ReadFile(name string) ([]byte, error) {
+	return os.ReadFile(name)
+}
+
+func (OsFs) ReadDir(dirname string) ([]fs.DirEntry, error) {
```
mpictor, line 128: Maybe add a pair of files, one for go versions before io/fs existed, defining DirEntry, and a second for newer versions that defines it as an alias?

Foxboron: Yes, I saw that was done for a few other libraries after I submitted the patch :)

Foxboron:
@bep Is there an intention of bringing the API of afero on par with the recent new functions in Go 1.16 and beyond?

PR #329, opened 2021-11-14
bugfix: RemoveAll

yearnfar:
when i RemoveAll lib directory, it also Removed libs directory

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -279,7 +279,7 @@ func (m *MemMapFs) RemoveAll(path string) error {
 	defer m.mu.RUnlock()
 
 	for p := range m.getData() {
-		if strings.HasPrefix(p, path) {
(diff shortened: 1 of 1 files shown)


jxsl13:
a test would be nice
and you need to press on the `cla bot signed yet` button.

mbertschler:
@yearnfar I don't fully understand what the problem is, you want the directory to be also removed but it currently isn't?

Can you please add a test for this problem?

jxsl13:
you add a trailing filepath separator in order to prevent other directories starting with the same prefix from being deleted as well as the one you actually want to delete.

General thought:
A lot of this stuff should not really rely on strings.HasPrefix but rather `filepath.Rel(pathA,pathB string) (relPath string, err error)`

For MemMapFs this should suffice as iirc it does add a trailing slash at the end of a directory name.

PR #339, opened 2022-02-06
Handle errors from closing writable Files

Jille:
Close might do some flushing, so failures to Close might cause data loss.

We should probably also check all Close() calls of read-only files, but I'm not aware of any data loss possibilities by ignoring those.


spf13:
Applied the fix to master directly using errors.Join rather than the errchain dependency. The core insight of your PR is correct — thanks for raising it.

PR #340, opened 2022-02-22
Fix gcsfs RemoveAll

typerat:
`RemoveAll` in `gcsfs` returns an error when the file/directory to be removed doesn't exist. It should return nil in this case.


mbertschler:
Thanks for your first contribution @typerat

Issue #342, opened 2022-03-01
[RFC] add Lchown


jxsl13:
Changing the owner of symlinks seems pretty useful to me.

I'd suggest the following method signature, same as LstatIfPossible.
```Go
type LinkOwner interface {
    LchownIfPossible(path string, uid, gid int) (bool, error)
}
``` 
The `bool` return value is `true` in case `Lchown` was used and `false` otherwise
This interface will most likely somehow require the `Lstater` interface to be implemented as well, as it's easiest to check whether the target path is a symlink and then decide based on that information whethe rto use `LchownIfPossible` or simply `Chown`

The `Lstater` interface implementation is not necessary for the `OsFs` but might be useful for other purposes down the line.

My use case is:
- Installer that runs as root in order to create a software setup
- Daemon that runs as technical user that should own all of the software written by the installer.

Discussion topics:
- Is this interface added to the Symlinker interface?
- Special error types needed?
-  ...

Looking at the other symlink related methods, the following interface seems to be the better option in addition to a custom PathError that is returned in case that the underlying implementation does not support this interface.

```go
type LinkOwner interface {
    LchownIfPossible(path string, uid, gid int) error
}
```

jxsl13:
Looking at the other symlink related methods, the following interface seems to be the better option in addition to a custom PathError that is returned in case that the underlying implementation does not support this interface.


```go
type LinkOwner interface {
    LchownIfPossible(path string, uid, gid int) error
}
```

PR #343, opened 2022-03-01
add LinkOwner interface #342

jxsl13:
Adds the ability to change ownership of symlinks.

read the issue at: #342

The use case is described in the issue as well.

Extends the `Symlinker` interface, requires more methods to be implemented.
Might be a *breaking change* in that regard which could be resolved by removing the `LinkOwner` interface from the `Symlinker` interface.

The change may be less severe due to most structs that implement parts of the symlinker interface have also been updated.

Feedback required :)


jxsl13:
Extends the `Symlinker` interface, requires more methods to be implemented.
Might be a breaking change in that regard.
The change may be less severe due to most structs that implement parts of the symlinker interface have also been updated.

Feedback required :)

jxsl13:
@mbertschler, @kargakis feedback pls :)
I do actively use this interface in order to keep track of filesystem modification in order to roll them back them if needed. This extension is a missing piece, especially when extracting tar archives containing symlinks as well as changing ownership of those to technical users.

PR #348, opened 2022-04-11
sftpfs: implement Symlinker and RemoveAll

luhuaei:


spf13:
Applied to master. These were clearly missing — thanks for the contribution.

PR #349, opened 2022-04-18
mem: add rudimentary (*FileInfo).Sys() method

dmke:
This allows downstream users to act on file ownership.

(A previous version added `mem.GetUID()`/`mem.GetGID()` methods, which didn't prove fruitful.)


spf13:
Applied to master. Useful improvement for testing permission-sensitive code.

PR #353, opened 2022-05-10
Fixes io.ReaderAt implementation of mem.File

### LINKED ISSUE #352: bug: `*mem.File` violates `io.ReaderAt` contract
The current implementation of `*mem.File` violates the contract defined by the Standard Library's `io.ReaderAt` interface.

Specifically, according to [the docs for `io.ReaderAt`](https://pkg.go.dev/io#ReaderAt):

> When ReadAt returns n < len(p), it returns a non-nil error explaining why more bytes were not returned. In this respect, ReadAt is stricter than Read.

The current implementation simply delegates `ReadAt` to `Read`:

https://github.com/spf13/afero/blob/100c9a6d7b9b70ddb10dc728b9b22d472e1f7f71/mem/file.go#L208-L214

This causes ReadAt to return a nil error when `ReadAt` reads less than the length of the give `[]byte`, while the contract requires it to return a non-nil error.

A secondary bug, consequence of the problem above, is that using the `fstest.TestFS(...)` func of the standard library fails if the tested files have any content. This happens because it delegates testing `ReadAt` of the given `fs.FS` of the test to `iotest.TestReader(...)`, which, in turn, makes an assertion on the condition described above.

In practical terms, I found this issue when I introduced `afero` and specificaly the `MemMapFs` to make some assertions on code I was writing that's supposed to generate and write files, and used `fstest.TestFS(...)` to check for the presence of the expected files I expected my code to generate.

It's an easy fix. Will soon send a PR with tests for both `*mem.File` and `NewIOFS` (i.e., `use MemMapFs`), and the fix.

bfreis:
Adds tests:
- `*mem.File` implementation of `io.ReaderAt` must return non-nil error when reading less than length of give slice.
- `MemMapFs` must pass `fstest.TestFS` tests for non-empty files.

Fixes #352 .

### DIFF
--- a/iofs_test.go
+++ b/iofs_test.go
@@ -40,7 +40,15 @@ func TestIOFS(t *testing.T) {
 			t.Fatal("OpenFile (O_CREATE) failed:", err)
 		}
 
-		f.Close()
+		_, err = f.WriteString("a")
+		if err != nil {
+			t.Fatal("WriteString failed:", err)
+		}
+
+		err = f.Close()
+		if err != nil {
+			t.Fatal("Close failed:", err)
+		}
 
 		if err := fstest.TestFS(NewIOFS(mmfs), "dir1/dir2/test.txt"); err != nil {
 			t.Error(err)
--- a/mem/file.go
+++ b/mem/file.go
@@ -223,6 +223,9 @@ func (f *File) ReadAt(b []byte, off int64) (n int, err error) {
 		n = len(f.fileData.data) - int(off)
 	}
 	copy(b, f.fileData.data[off:off+int64(n)])
+	if n < len(b) {
+		err = io.EOF
+	}
 	return
 }
 
@@ -231,7 +234,6 @@ func (f *File) Read(b []byte) (n int, err error) {
 	atomic.AddInt64(&f.at, int64(n))
 	return
 }
-
 func (f *File) Truncate(size int64) error {
 	if f.closed {
 		return ErrFileClosed
--- a/mem/file.go
+++ b/mem/file.go
@@ -232,6 +232,9 @@ func (f *File) ReadAt(b []byte, off int64) (n int, err error) {
 func (f *File) Read(b []byte) (n int, err error) {
 	n, err = f.ReadAt(b, atomic.LoadInt64(&f.at))
 	atomic.AddInt64(&f.at, int64(n))
+	if n > 0 && err == io.EOF {
+		err = nil
+	}
 	return
 }
 func (f *File) Truncate(size int64) error {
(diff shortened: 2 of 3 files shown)


bfreis:
👋 @mbertschler, @kargakis

Hey folks, pinging you as I see you merged some PRs recently :)

I have some code commented in tests just waiting for this to get merged. Thanks!

spf13:
The conformance fix is correct — `io.ReaderAt` requires a non-nil error when `n < len(b)`, and the current `Read` delegation violates that. The `fstest.TestFS` failure confirms it.

However, the implementation has two problems:

**1. Data race.** After `f.Read(b)` releases the lock, the fix accesses `f.fileData.data` without holding it:

```go
n, err = f.Read(b)                                        // lock released here
atomic.StoreInt64(&f.at, prev)
if int64(n) == int64(len(f.fileData.data))-off {          // ← unguarded access
```

A concurrent write can resize `f.fileData.data` between those lines. The race detector will catch this.

**2. `ReadAt` shouldn't delegate to `Read` at all.** The `io.ReaderAt` contract says ReadAt must not affect nor be affected by the file's seek offset. The current implementation manipulates `f.at` with atomics to fake independence, but that's not safe under concurrent `Read`/`ReadAt` calls. This is a pre-existing issue, but the fix shouldn't deepen it.

The correct implementation:

```go
func (f *File) ReadAt(b []byte, off int64) (n int, err error) {
    f.fileData.Lock()
    defer f.fileData.Unlock()
    if f.closed {
        return 0, ErrFileClosed
    }
    if off >= int64(len(f.fileData.data)) {
        return 0, io.EOF
    }
    n = copy(b, f.fileData.data[off:])
    if n < len(b) {
        err = io.EOF
    }
    return
}
```

This holds the lock for the full operation, doesn't touch `f.at`, is race-free, and satisfies the `io.ReaderAt` contract including the EOF requirement. Also note that #448 touches this same function — worth coordinating.

The test changes and the `iofs_test.go` fix are good as-is.

spf13:
Update: #448 was just merged, which correctly restructures `ReadAt` as the primary implementation (holds the lock, operates on `off` directly, does not touch `f.at`). That resolves the race and the seek-offset concerns raised above.

The remaining gap is exactly what this PR targets: `ReadAt` still returns `n < len(b), nil` instead of `n < len(b), io.EOF` when the read hits end of file. The fix now becomes trivial against the new code — just add before the return in `ReadAt`:

```go
if n < len(b) {
    err = io.EOF
}
```

Please rebase onto master and update the implementation accordingly. The test changes here are still valid and needed.

Issue #361, opened 2022-06-16
MemMapFS: missing files after write multi files in concurrency


sko00o:
I have used `afero.MemMapFs` for filesystem mock test in my projects.
But I got some unexpected test results.
 
I want to list all files in target directory that I have wrote multi files in concurrency before.
Occasionally some files are not returned by `Readdir` function. Also I had the same problem calling `Walk` function on this directory.

Here is a test to reproduce my problem. https://github.com/sko00o/afero-test/blob/main/memmap/memmap_test.go#L42

failure result:

```
❯ go test -race -run ^TestListFilesAfterConcurrencyFileWrite$ ./... -test.count 1
--- FAIL: TestListFilesAfterConcurrencyFileWrite (4.39s)
    memmap_test.go:99: 
                Error Trace:    memmap_test.go:99
                Error:          Not equal: 
                                expected: 10000
                                actual  : 9995
                Test:           TestListFilesAfterConcurrencyFileWrite
                Messages:       missing files
FAIL
FAIL    github.com/sko00o/afero-test/memmap     4.677s
FAIL
```

----
I notice another issue #298 that found `MemMapFs.Mkdir` have problem when call in concurrency.
In my opinion, we need a double check [here](https://github.com/spf13/afero/blob/v1.8.2/memmap.go#L145) before we create new directory.

----
So I send a PR for this.

sko00o:
Ping @bep 

Please check this issue.

For quick reproduce: https://go.dev/play/p/3xkOSEBVoNS

bep:
Your failing test case looks correct to me (as in, it should not fail), which comes a little as a surprise to mek as I use this pretty concurrently myself ... It's not obvious to me why, though.

I also agree about the Mkdir bug, assuming that that's how most sane real file systems work?

PR #362, opened 2022-06-16
Fix memmap mkdir

sko00o:
Fix for mkdir in concurrency.
Related issue https://github.com/spf13/afero/issues/361 .

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -153,7 +153,7 @@ func (m *MemMapFs) Mkdir(name string, perm os.FileMode) error {
 	m.registerWithParent(item, perm)
 	m.mu.Unlock()
 
-	return m.setFileMode(name, perm|os.ModeDir)
+	return nil
 }
 
 func (m *MemMapFs) MkdirAll(path string, perm os.FileMode) error {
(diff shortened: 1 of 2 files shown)


memmap.go
```diff
@@ -142,13 +142,18 @@ func (m *MemMapFs) Mkdir(name string, perm os.FileMode) error {
 	}
 
 	m.mu.Lock()
+	// double check before create new dir
+	if _, ok := m.getData()[name]; ok {
+		m.mu.Unlock()
+		return &os.PathError{Op: "mkdir", Path: name, Err: ErrFileExists}
+	}
```
acabarbaye, line 149: I would `defer` the `unlock` line 145 so that it happens no matter the path of the code chosen and there is no risk of stale locks
```suggestion
      defer m.mu.Unlock()
	// double check before create new dir
	if _, ok := m.getData()[name]; ok {
		return &os.PathError{Op: "mkdir", Path: name, Err: ErrFileExists}
	}
```

sko00o: LGTM.

sko00o:
Already fixed in a800a9d. Okey, I will close this PR.

Issue #373, opened 2022-07-25
MemMapFS creates files even when directories do no exist


natebrennand:
I intended to use MemMapFS to fully test my filesystem access in an application, but accidentally introduced a bug where I was writing to a file before creating the directory.

Is this mismatch in behavior intentional?

k3an3:
Ran into this issue as well; PoC:

```go
package idk

import (
	"errors"
	"io/fs"
	"os"
	"path/filepath"
	"testing"

	"github.com/spf13/afero"
)

func TestOsFs(t *testing.T) {
	appFS := afero.NewOsFs()
	testPath1 := filepath.Join("/tmp", "test")
	f, err := appFS.Create(testPath1)
	if err != nil {
		t.Fatal(err)
	}
	f.Close()

	testPath2 := filepath.Join("/tmp", "test", "test")
	_, err = appFS.Create(testPath2)
	var pathErr *fs.PathError
	if !errors.As(err, &pathErr) {
		t.Fatalf("Expected file open of `%s' to fail: err was %v", testPath2, err)
	}
}

func TestMemMapFs(t *testing.T) {
	appFS := afero.NewMemMapFs()
	testPath1 := filepath.Join("/tmp", "test")
	f, err := appFS.Create(testPath1)
	if err != nil {
		t.Fatal(err)
	}
	f.Close()

	testPath2 := filepath.Join("/tmp", "test", "test")
	_, err = appFS.Create(testPath2)
	var pathErr *fs.PathError
	if !errors.As(err, &pathErr) {
		t.Fatalf("Expected file open of `%s' to fail: err was %v", testPath2, err)
	}
}
```

I would expect `afero.MemMapFs` and `afero.OsFs` to behave the same way in this regard.

PR #374, opened 2022-07-25
add option to have strict file creation behavior

natebrennand:
Workaround for https://github.com/spf13/afero/issues/373

Adds an option to have MemMapFS fail when creating a file in a directory that DNE.

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/spf13/afero/mem"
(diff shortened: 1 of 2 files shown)


Frassle:
I'd consider it a bug that it ever allowed this in the first place. A file create on `OsFs` would fail, it would be less surprising if `MemFs` also had that behavior. We've had bugs slip by testing due to this difference, but I'm not sure we would have noticed that we need to turn strict mode on for `MemFs` if it was an option like this.

natebrennand:
I agree that it's a bug, but fixing it would be a breaking change.
Dunno that a v2 is anywhere on the horizon

Frassle:
> fixing it would be a breaking change.

Obligatory xkcd: https://xkcd.com/1172/

k3an3:
Any updates on whether this will be merged? I can't imagine this bug would be expected behavior anywhere.

spf13:
Closing this. The correct fix for #373 is making MemMapFs match POSIX open() semantics by default — which is what #572 proposes. Adding a permanent opt-in flag for correct behavior is the wrong direction: it creates two behavioral modes, adds API surface that becomes dead code if #572 is ever merged, and requires callers to explicitly ask for semantics that should just be correct out of the box. Closing in favor of #572.

PR #377, opened 2022-08-22
Rclone

mkvolkov:
Added Rclone virtual filesystem with the full functionality of afero and afero.Fs functions available in rclone.


jxsl13:
Hi, cool fs abstraction.
You should take a closer look at other implementation in regard to what kind of errors are returned in case some methods are not supported or in case operations fail.

Usually os.PathError or os.LinkError are used.
tarfs/zipfs might be good references.

mkvolkov:
@kargakis please take a look

mkvolkov:
@bep take a look please

mkvolkov:
@mbertschler @kargakis @bep where is everybody?

bep:
@mkvolkov that is a good question, I can only answer for myself; I'm rather busy these days, and I often need to put a priority on the repositories that I consider "mine". I'm a big fan of Rclone, though.

spf13:
Thank you for building this. Afero's design philosophy is to be a thin, composable abstraction layer - the core package ships a small set of general-purpose filesystem types and a handful of well-maintained backend drivers. We've decided not to expand that set with additional external-service backends inside this repo, because each one adds a permanent maintenance burden, external service dependency, and CI complexity.

The right home for this is a standalone module published under your own namespace (e.g. github.com/mkvolkov/afero-rclone for Rclone, github.com/kolaente/afero-swift for Swift, etc.). That gives you full control over releases and dependencies, and makes it immediately usable without pulling additional dependencies into afero itself.

We'd be happy to list community backends in the README. If you publish it as a standalone module, open a small PR adding a line to the README and we'll merge it promptly.

PR #389, opened 2023-04-05
Fix MemMapFs inconsistent path store

Frassle:
Adds a test that fails for the current MemMapFs when creating files in nested directories with synonyms for the path (i.e. full path, relative path, explict relative path).

Also fixes MemMapFs so that test (and all others pass).


spf13:
Applied to master. Clear correctness issue — sorry it sat so long.

PR #393, opened 2023-07-10
Add check for path separator to TempDir

leonrosenshein:
Since GO 1.17, ioutil.TempDir has called MkdirTemp for it's implementation. Since at least October of 2020, MkdirTemp has [checked the input prefix for os.PathSeparator](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/os/tempfile.go;l=62;bpv=0;bpt=0). This PR adds that check and returns an error in that case.

The PR also includes some basic tests for the TempDir function.

### DIFF
--- a/ioutil.go
+++ b/ioutil.go
@@ -17,6 +17,7 @@ package afero
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


spf13:
Applied to master. Thanks for catching this.

PR #398, opened 2023-08-07
Added dependabot

Neo2308:
* Added dependabot to keep go dependencies & actions updated.


Neo2308:
@bep could you review this?

PR #406, opened 2023-10-17
Add network interface for minio

cpyun:
Add network interface for [minio](https://min.io)


cpyun:
This also applies to AWS S3

PR #409, opened 2023-11-14
Fix the infinite recursion with afero.Walk

### LINKED ISSUE #185: MemMapFs + afero.Walk -> infinite loop
I have an MemMapFs with the following structure:
```go
([]os.FileInfo) (len=1 cap=1) {
 (*mem.FileInfo)(0xc420098080)({
  FileData: (*mem.FileData)(0xc4200a6300)({
   Mutex: (sync.Mutex) {
    state: (int32) 0,
    sema: (uint32) 0
   },
   name: (string) (len=44) "abc.txt_a3171851-d16d-4b27-a790-1a4d3d4f221a",
   data: ([]uint8) (len=11 cap=16) {
    00000000  48 65 6c 6c 6f 20 47 6f  6c 70 21                 |Hello Golp!|
   },
   memDir: (mem.Dir) <nil>,
   dir: (bool) false,
   mode: (os.FileMode) T---------,
   modtime: (time.Time) 2018-12-01 09:58:08.052452176 +0100 CET m=+0.001122979
  })
 })
}
```
(One file named: `abc.txt_a3171851-d16d-4b27-a790-1a4d3d4f221a`)

So it should only call the `walkFn` once.

```go
afero.Walk(i.Fs(), "/", func(path string, info os.FileInfo, err error) error {
	spew.Dump(info)
	return nil
})
```

But the above code results in:
```go
(*mem.FileInfo)(0xc42000e088)({
 FileData: (*mem.FileData)(0xc420068300)({
  Mutex: (sync.Mutex) {
   state: (int32) 0,
   sema: (uint32) 0
  },
  name: (string) (len=1) "/",
  data: ([]uint8) <nil>,
  memDir: (*mem.DirMap)(0xc42000e078)((len=1) {
   (string) (len=44) "abc.txt_2d83e30d-bf11-4953-8ebb-fdb2e91235cc": (*mem.FileData)(0xc420068360)({
    Mutex: (sync.Mutex) {
     state: (int32) 0,
     sema: (uint32) 0
    },
    name: (string) (len=44) "abc.txt_2d83e30d-bf11-4953-8ebb-fdb2e91235cc",
    data: ([]uint8) (len=11 cap=16) {
     00000000  48 65 6c 6c 6f 20 47 6f  6c 70 21                 |Hello Golp!|
    },
    memDir: (mem.Dir) <nil>,
    dir: (bool) false,
    mode: (os.FileMode) T---------,
    modtime: (time.Time) 2018-12-01 09:53:37.422059676 +0100 CET m=+0.001065426
   })
  }),
  dir: (bool) true,
  mode: (os.FileMode) ----------,
  modtime: (time.Time) 0001-01-01 00:00:00 +0000 UTC
 })
})
(*mem.FileInfo)(0xc42000e100)({
 FileData: (*mem.FileData)(0xc420068600)({
  Mutex: (sync.Mutex) {
   state: (int32) 0,
   sema: (uint32) 0
  },
  name: (string) (len=1) "/",
  data: ([]uint8) <nil>,
  memDir: (*mem.DirMap)(0xc42000e0b0)((len=1) {
   (string) (len=1) "/": (*mem.FileData)(0xc420068600)(<already shown>)
  }),
  dir: (bool) true,
  mode: (os.FileMode) T---------,
  modtime: (time.Time) 2018-12-01 09:53:37.424199616 +0100 CET m=+0.003205398
 })
})
(*mem.FileInfo)(0xc420174010)({
 FileData: (*mem.FileData)(0xc420068720)({
  Mutex: (sync.Mutex) {
   state: (int32) 0,
   sema: (uint32) 0
  },
  name: (string) (len=1) "/",
  data: ([]uint8) <nil>,
  memDir: (*mem.DirMap)(0xc42000e110)((len=1) {
   (string) (len=1) "/": (*mem.FileData)(0xc420068720)(<already shown>)
  }),
  dir: (bool) true,
  mode: (os.FileMode) T---------,
  modtime: (time.Time) 2018-12-01 09:53:37.425307087 +0100 CET m=+0.004312958
 })
})
(*mem.FileInfo)(0xc420174070)({
 FileData: (*mem.FileData)(0xc42017e0c0)({
  Mutex: (sync.Mutex) {
   state: (int32) 0,
   sema: (uint32) 0
  },
  name: (string) (len=1) "/",
  data: ([]uint8) <nil>,
  memDir: (*mem.DirMap)(0xc420174020)((len=1) {
   (string) (len=1) "/": (*mem.FileData)(0xc42017e0c0)(<already shown>)
  }),
  dir: (bool) true,
  mode: (os.FileMode) T---------,
  modtime: (time.Time) 2018-12-01 09:53:37.426633396 +0100 CET m=+0.005639164
 })
})
// It goes on forever
```
So the `walkFn` is called once for the file and an infinite amount of times for `/`.
(It doesn't work with `.` instead of `/` either)

erstam:
This PR fixes https://github.com/spf13/afero/issues/185

Fix the infinite recursion by skipping the current directory in the directory names listing

### DIFF
--- a/afero_test.go
+++ b/afero_test.go
@@ -32,7 +32,7 @@ var (
 	Fss      = []Fs{&MemMapFs{}, &OsFs{}}
 )
 
-var testRegistry map[Fs][]string = make(map[Fs][]string)
+var testRegistry = make(map[Fs][]string)
 
 func testDir(fs Fs) string {
 	name, err := TempDir(fs, "", "afero")
--- a/path.go
+++ b/path.go
@@ -59,8 +59,8 @@ func walk(fs Fs, path string, info os.FileInfo, walkFn filepath.WalkFunc) error 
 	}
 
 	for _, name := range names {
-		if name == path {
-			// skip current directory to avoid infinite recursion
+		if name == "" {
+			// skip empty names to avoid infinite recursion
 			continue
 		}
 		filename := filepath.Join(path, name)
--- a/path.go
+++ b/path.go
@@ -59,6 +59,10 @@ func walk(fs Fs, path string, info os.FileInfo, walkFn filepath.WalkFunc) error 
 	}
 
 	for _, name := range names {
+		if name == path {
+			// skip current directory to avoid infinite recursion
+			continue
+		}
 		filename := filepath.Join(path, name)
 		fileInfo, err := lstatIfPossible(fs, filename)
 		if err != nil {
--- a/path_test.go
+++ b/path_test.go
@@ -67,3 +67,27 @@ func TestWalk(t *testing.T) {
 		t.Fail()
 	}
 }
+
+func TestWalkRemoveRoot(t *testing.T) {
+	defer removeAllTestFiles(t)
+	fs := Fss[0]
+	rootPath := "."
+
+	err := fs.RemoveAll(rootPath)
+	if err != nil {
+		t.Error(err)
+	}
+
+	testRegistry[fs] = append(testRegistry[fs], rootPath)
+	setupTestFiles(t, fs, rootPath)
+
+	walkFn := func(path string, info os.FileInfo, err error) error {
+		fmt.Println(path, info.Name(), info.IsDir(), info.Size(), err)
+		return err
+	}
+
+	err = Walk(fs, rootPath, walkFn)
+	if err != nil {
+		t.Error(err)
+	}
+}


erstam:
@spf13 Please, who can review this PR ?

erstam:
@bep  I worked a bit more on this issue and noticed something interesting.  In my code I am removing the root folder of the fs instance and this causes the infinite recursion when using the Walk method.  Should removing the root be forbidden on all Fs types ?  I see in MemMapFs there is a `init sync.Once` attribute which is used to initialize a "root path" only once.  But what if we delete that root path ?  It kinda breaks the Fs.

The below test function will reproduce the issue.

```go
func TestWalkRemoveRoot(t *testing.T) {
	defer removeAllTestFiles(t)
	fs := Fss[0]
	rootPath := "."

	err := fs.RemoveAll(rootPath)
	if err != nil {
		t.Error(err)
	}

	testRegistry[fs] = append(testRegistry[fs], rootPath)
	setupTestFiles(t, fs, rootPath)

	walkFn := func(path string, info os.FileInfo, err error) error {
		fmt.Println(path, info.Name(), info.IsDir(), info.Size(), err)
		return err
	}

	err = Walk(fs, rootPath, walkFn)
	if err != nil {
		t.Error(err)
	}
}
```

Bases from this case, my proposed fix is just a workaround.  The real fix would be to better protect the root path from deletion.

erstam:
@bep, @spf13 any thoughts on my previous comment ?

erstam:
Will we see this merged at some point ?

spf13:
Applied and merged directly to master with a cleaned-up test (removed fmt.Println from the test callback). Thank you for tracking this down!

PR #416, opened 2024-02-07
Fix O_EXCL without O_CREATE in memmap fs

jimsnab:
Fix issue where OpenFile fails incorrectly trying to open an existing file with O_EXCL in memmap fs.


spf13:
@jimsnab the fix here looks good and we'd like to merge it, but the CLA check is still pending. Please sign at https://cla-assistant.io/spf13/afero — it only takes a moment. If we don't hear back in two weeks we'll close this and someone can reopen once the CLA is signed.

PR #417, opened 2024-02-22
Add new relpath filesystem

purpleidea:
This new filesystem implements a relative filesystem modifier which can
be useful for converting between absolute filesystems rooted at / and
relative ones. This is particularly useful when interfacing with the
golang embed package.

### DIFF
--- a/relpath.go
+++ b/relpath.go
@@ -1,0 +1,216 @@
+// Copyright (C) 2024+ James Shubin
+// Written by James Shubin <james@shubin.ca>
+// This filesystem implementation is based on the afero.BasePathFs code.
+package afero
+
+import (
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+var (
+	_ Lstater        = (*RelPathFs)(nil)
+	_ fs.ReadDirFile = (*RelPathFile)(nil)
+	_ fs.ReadDirFile = readDirFile{}
+)
+
+// RelPathFs removes a prefix from all operations to a given path within an Fs.
+// The given file name to the operations on this Fs will have a prefix removed
+// before calling the base Fs.
+//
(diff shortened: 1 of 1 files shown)


purpleidea:
Sorry CLA's are a rude form of asymmetrical power against contributors who are giving you something for free. There's no good reason for a CLA here.

CrimsonFez:
> Sorry CLA's are a rude form of asymmetrical power against contributors who are giving you something for free. There's no good reason for a CLA here.

Did you actually read it??? They just want you to verify that you have the rights to the code you're submitting.

spf13:
Two issues before this can move forward:

1. **CLA not signed.** Please sign the CLA at https://cla-assistant.io/spf13/afero — required for all contributions.

2. **`TrimPrefix` without `HasPrefix` guard.** `RealPath` calls `strings.TrimPrefix(name, r.prefix)` unconditionally. If a caller passes a path that doesn't start with the configured prefix, the path is returned unchanged rather than an error being returned. There's a commented-out guard in your code suggesting you were aware of this — it should be enforced. Something like:
   ```go
   if !strings.HasPrefix(name, r.prefix) {
       return name, os.ErrNotExist
   }
   ```

3. **No tests.** `RelPathFs` needs at least a table-driven test covering: paths that have the prefix (should strip), paths that don't have the prefix (should error), and `Name()` on a file (should add prefix back).

The concept is valid and distinct from `BasePathFs` (which prepends; this strips). Fix the above and it's a reasonable addition.

spf13:
Closing with thanks — the use case you identified (bridging between absolute and relative path spaces, particularly for embed.FS) is real and valid. However, this is now better addressed by the NewBridgeIOFS adapter (#575), which lets you import any io/fs.FS directly as an afero.Fs. That's the direction the project is taking for this problem. Thanks for identifying the gap early.

PR #420, opened 2024-03-25
allow absolute paths in BasePathFs when prefix is "" on windows

DenKoren:
As a special case, it is OK to allow absolute paths access when the `BasePathFs.path` is empty.
Sometimes we need to provide access to full FS through BasePathFs type for code simplicity (to pass it to the functions, that want `BasePathFs` specific type, not just `afero.FS` interface).


spf13:
Closing this. A BasePathFs with an empty path is a degenerate wrapper that restricts nothing, which makes it harder to reason about the invariant that BasePathFs always confines access to a subtree. There's no API in afero that requires *BasePathFs specifically (everything takes afero.Fs), so the use case of passing a full FS as a BasePathFs-typed argument doesn't arise in practice. If you need a BasePathFs that allows access to the entire filesystem, `NewBasePathFs(fs, "/")` is the right idiom.

PR #424, opened 2024-04-29
Add buckets caching to GcsFS

meandnano:
Without it each files interaction on GCS results in an unnecessary GetBucket request.


spf13:
Closing — superseded by #568, which is a strict superset: same bucket caching, plus a proper `sync.RWMutex` protecting the cache map (avoiding the data race your PR would have introduced with concurrent access), plus configurable upload chunk size. Thanks for identifying the missing initialization — that was a real bug.

PR #429, opened 2024-07-28
Add sftpfs files WriteAt and Sync methods

MRtecno98:
Similarly to #378 the WriteAt and Sync methods are also missing and a TODO despite being implemented by sftp.File already.

Also may i suggest just nesting sftp.File into sftpfs.File to have all necessary interfaces implemented without delegate methods? Something like this

```
type File struct {
        *sftp.File
	client *sftp.Client
}
```

This would allow removing methods `Close Name Stat Sync Truncate Read ReadAt Seek Write WriteAt` which are just simple delegates. I'll gladly implement the change if it's ok otherwise you can just merge this pr as it is.

### DIFF
--- a/sftpfs/file.go
+++ b/sftpfs/file.go
@@ -100,9 +100,8 @@ func (f *File) Write(b []byte) (n int, err error) {
 	return f.fd.Write(b)
 }
 
-// TODO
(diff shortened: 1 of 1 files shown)


spf13:
Applied to master. Thanks for tracking these down!

PR #437, opened 2025-01-06
Use PosixRename for sftpfs

jonathan-conder:
OpenSSH 4.8 added POSIX-style renames as a protocol extension [1]. The sftp library supports the extension, on both the client and server side, since version 1.13 [2].

For consistency with the other filesystems, it makes sense to use POSIX renames for sftpfs. The old-style rename does not overwrite existing files, which makes it difficult to implement atomic file writes.

[1] https://www.openssh.com/txt/release-4.8
[2] https://github.com/pkg/sftp/pull/384

### DIFF
--- a/sftpfs/sftp.go
+++ b/sftpfs/sftp.go
@@ -116,7 +116,7 @@ func (s Fs) RemoveAll(path string) error {
 }
 
 func (s Fs) Rename(oldname, newname string) error {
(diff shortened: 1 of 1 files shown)


spf13:
Applied to master. PosixRename is the right call here — thanks for the fix.

Issue #433, opened 2024-11-20
`mem.ReadAt/`mem.WriteAt` violates spec and subject to race conditions


k3an3:
Note in the Go 1.23.3 [docs](https://pkg.go.dev/io#WriterAt) for the `io.WriterAt` interface:

> If WriteAt is writing to a destination with a seek offset, WriteAt should not affect nor be affected by the underlying seek offset.

> Clients of WriteAt can execute parallel WriteAt calls on the same destination if the ranges do not overlap.

However, the current implementations of MemFS ReadAt/WriteAt look like this, respectively:

https://github.com/spf13/afero/blob/5c4385aa20510dba5ca4db12c02b0c9211d82892/mem/file.go#L226-L232

https://github.com/spf13/afero/blob/5c4385aa20510dba5ca4db12c02b0c9211d82892/mem/file.go#L300-L303

If multiple calls are made in parallel on the same file, a race will occur between modifying the stored file offset and reading/writing the file, causing these functions to return or write incorrect data. Additionally, it is not expected that the underlying seek offset is modified at all. A corrected implementation will likely have to extract and re-implement parts of Read/Write in order to not affect the saved offset.

PR #448, opened 2025-01-28
Fix ReadAt + WriteAt concurrency

royiarchy:
Fixing https://github.com/spf13/afero/issues/433 to allow parallel executions of `ReadAt` and `WriteAt` without messing up the seek offset.

### DIFF
--- a/mem/file.go
+++ b/mem/file.go
@@ -201,33 +201,30 @@ func (f *File) ReadDir(n int) ([]fs.DirEntry, error) {
 	return di, nil
 }
 
-func (f *File) Read(b []byte) (n int, err error) {
+func (f *File) ReadAt(b []byte, off int64) (n int, err error) {
 	f.fileData.Lock()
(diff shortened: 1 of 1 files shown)


spf13:
Merging this — the structural fix is correct. Inverting the delegation so `ReadAt`/`WriteAt` are the primary implementations (holding the lock, operating directly on `off`, not touching `f.at`) is exactly right.

One gap: the new `ReadAt` still does not return `io.EOF` when `n < len(b)` and the read hits end of file. The `io.ReaderAt` contract requires a non-nil error in that case. Example: 4-byte file, `off=0`, `len(b)=5` → returns `4, nil` instead of `4, io.EOF`. `fstest.TestFS` will catch this.

The fix is simple — add before the return in `ReadAt`:

```go
if n < len(b) {
    err = io.EOF
}
```

Tracking that in #353, which will need to be updated to apply against this new implementation.

PR #462, opened 2025-02-18
Split gcsfs and sftpfs into separate modules

sagikazarmark:
This PR splits the library into three modules:

- the main module contains all code without external dependencies
- gcsfs and sftpfs are FS implementations with external dependencies

This change is modeled after a similar, [recent change in Viper](https://github.com/spf13/viper/pull/1860) to reduce dependencies.

@bep @spf13 let me know what you think.

## Todo

- [x] Make sure to run tests for submodules


bep:
I think this is a great idea.

bep (approved):

spf13:
Y'all move fast. I think this is a brilliant idea. Something we couldn't have easily done without modules, but now with modules, it's a no brainer to do this.

Issue #204, opened 2019-03-26
UnionFile.Readdir checks for c != -1, not <= 0


bep:
Which isn't in line with the `os` specification:

    If n <= 0, Readdir returns all the FileInfo from the directory in a single slice. 

See https://golang.org/pkg/os/#File.Readdir

amitmishra11:
This appears to already be fixed on `master` — `UnionFile.Readdir` checks `if c <= 0` (not `c != -1`), matching the `os.File.Readdir` spec quoted in the issue. Fixed by 588a75e ("Fix UnionFile.Readdir return when c <= 0") and 066c8b0 ("Fix amount of files read in UnionFile.Readdir"), via #260. Looks safe to close.

PR #481, opened 2025-03-21
fix: UnionFile.Readdir now returns all entries when count <= 0 (fixes…

ljluestc:
… #204)


spf13:
Closing this one. The PR renames parameter \c\ to \count\ but makes zero behavioral change — the original code already used \c <= 0\, not \c != -1\ as the added comment claims. The comment is factually incorrect and would mislead future readers. If issue #204 describes a real bug, a PR with an actual fix is welcome.

Issue #299, opened 2021-04-10
can add githubfs?


runner-mei:
can add githubfs?
github repo as a filesystem

PR #483, opened 2025-03-30
Add githubfs feature

ljluestc:
can add githubfs #299

### DIFF
--- a/basepath.go
+++ b/basepath.go
@@ -55,12 +55,16 @@ func (b *BasePathFs) RealPath(name string) (path string, err error) {
 		return name, err
 	}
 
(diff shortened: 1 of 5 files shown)


spf13:
Thank you for the idea. Afero's design philosophy is to keep the core package to a small set of general-purpose filesystem types and well-maintained backends. Adding a GitHub API backend inside this repo would add a permanent external API dependency and CI complexity.

The right home for this is a standalone module (e.g. github.com/ljluestc/afero-github). We'd be happy to list community backends in the README - if you publish it, open a small PR adding a line to the README and we'll merge it promptly.

PR #487, opened 2025-04-03
construct a new bytes.Reader per each opened file in tarfs

### LINKED ISSUE #486: tarfs.FS uses shared file position for each file handler
Playground: https://go.dev/play/p/kNiuE6X1qtk

When using `tarfs.FS`, if you open a file, read it, close it. The index for the file seeker is at the end of the file. So if you open it again later, any `read` will return an empty slice.

This is because the file handler returned is the same for both callers. All files are stored in a map, and returned as a struct copy to each caller: 

Cached files:

https://github.com/spf13/afero/blob/master/tarfs/fs.go#L17

File copy:

https://github.com/spf13/afero/blob/master/tarfs/fs.go#L83-L97

This struct copy of the `File` has its data in a `*bytes.Reader`, so all file handlers for a filepath share the same reader:

https://github.com/spf13/afero/blob/master/tarfs/file.go#L16

And `bytes.Reader` has an [index](https://github.com/golang/go/blob/master/src/bytes/reader.go#L20) for the current position:

```golang
type Reader struct {
	s        []byte
	i        int64 // current reading index
	prevRune int   // index of previous rune; or < 0
}
```

-----

File handles to the `tarfs.File` should all have their own unique file seek location. This was fixed for `MemMapFS` files a long time ago: https://github.com/spf13/afero/pull/55/files

I think `tarfs.File` needs a similar fix. Or just make it use `mem.File`

aslilac:
Closes #486

Currently, every call to `Open` "copies" the `File`, but since it has pointer fields, it's not really the kind of clean independent copy that's desired here. Notably, every `File` shares the same `bytes.Reader` state.

This causes a few bugs:
- A file which has been read to the end will be unreadable by anyone else.
- Any file which has been `Close`d has had `data` set to `nil`, essentially permanently erasing that file.

My proposed solution is to introduce an internal `fsEntry` type, and to construct a new `tarfs.File` from that data on each call to `Open`. Continuing to share a `*tar.Header` field seems fine, as you shouldn't modify that during normal use.

### DIFF
--- a/tarfs/fs.go
+++ b/tarfs/fs.go
@@ -14,7 +14,12 @@ import (
 )
 
 type Fs struct {
-	files map[string]map[string]*File
+	files map[string]map[string]*fsEntry
+}
+
+type fsEntry struct {
+	h    *tar.Header
+	data []byte
 }
 
 func splitpath(name string) (dir, file string) {
@@ -29,7 +34,7 @@ func splitpath(name string) (dir, file string) {
 }
 
 func New(t *tar.Reader) *Fs {
-	fs := &Fs{files: make(map[string]map[string]*File)}
+	fs := &Fs{files: make(map[string]map[string]*fsEntry)}
 	for {
 		hdr, err := t.Next()
 		if err == io.EOF {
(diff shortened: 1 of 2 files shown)


Emyrk:
@sagikazarmark are PRs being actively reviewed?

sagikazarmark:
Makes sense to me. Added to my review queue.

dannykopping:
@sagikazarmark following up on this one; do you intend to review/merge it any time soon?

spf13:
Applied to master. Genuine correctness bug — sorry it sat this long.

PR #507, opened 2025-05-25
463 support os root

sean9999:
add support for os.Root.
see: https://go.dev/blog/osroot.

This code mainly sticks to just the task at hand, however in order to make CI tests pass, I had to do some very minor clean-ups.

I introduce the interface `Root`, which satisfies the same method-set as os.Root.

Root is basically a "chroot". It offers a root-relative view into a filesystem that you can't escape.

Root is powered by *rootFs, which simply wraps *BasePathFs, changing the signature slightly so that it satisfies the interface, and the functionality that os.Root provides.


sean9999:
> Thanks @sean9999
> 
> I think this is a great addition!
> 
> If you don't mind, I'd like to wait for 1.25 to be released: we usually try to support two versions.

It's time, boss! =)
https://go.dev/blog/go1.25

spf13:
The concept is good — `os.Root` (Go 1.24) fits naturally with afero's layered FS model. But adding `OpenRoot` to the `Fs` interface is a breaking change: every external `afero.Fs` implementation breaks at compile time. That's the same problem as adding a method to `File`, and the fix is the same pattern:

```go
type RootOpener interface {
    OpenRoot(name string) (Root, error)
}
```

Callers type-assert. The built-in backends implement it where it makes sense.

Other blocking issues:
1. `go.mod` minimum bumped to Go 1.24 — this constrains all afero users. If `os.Root` is behind a `RootOpener` interface, the minimum version bump can stay local to callers who use it (or be gated behind a build tag).
2. testify added to the core `go.mod` — keep test-only dependencies in a separate `go.mod` or use stdlib `testing`/`testing/quick`.
3. CI must pass on current master before we can merge.

Please re-submit with the `RootOpener` interface approach, with CI green.

PR #529, opened 2025-08-18
feat: support hardlinks for unix os's

ivan1993spb:


spf13:
Two issues:

1. **Wrong build constraint.** `//go:build !windows` excludes Windows, but `os.Link` has worked on Windows since Go 1.16. Remove the build tag and put this directly in `os.go` (or an untagged `os_link.go`).

2. **No Linker interface.** Without a corresponding interface, `Link` is an undiscoverable method — callers using `afero.Fs` have no way to check for or use it. The established pattern in this repo is a narrow optional interface:

```go
// In symlink.go, alongside Symlinker:
type Linker interface {
    LinkIfPossible(oldname, newname string) error
}
```

(Using `LinkIfPossible` to match the naming convention of `SymlinkIfPossible`.)

Fix both and this is a straightforward merge.

PR #568, opened 2025-12-19
Add support for configurable GCS upload chunk size

jacobfederer:
This pull request enhances the GCS upload functionality by introducing a `SetUploadChunkSizeByte` configuration option, allowing users to define custom chunk sizes for uploads. This improvement provides greater flexibility and optimization for specific upload scenarios.


spf13:
Applied to master. The mutex fix alone was worth this PR — thanks for the thorough approach.

PR #572, opened 2026-02-23
fix: MemMapFs.OpenFile checks parent dir exists before creating file

### LINKED ISSUE #270: WriteFile creates non-existing directory for MemMapFs but not for OsFs
The code below creates the non-existing directory "nonexistingdir" for MemMapFs but returns an error for OsFs. Shouldn't the behavior be the same?

```go
package main

import (
	"github.com/spf13/afero"
	"log"
)

func main() {
	memFs := afero.NewMemMapFs()
	err := afero.WriteFile(memFs, "/tmp/nonexistingdir/testfile.txt", []byte("content"), 0777)
	if err != nil {
		log.Printf("write MemMapFs file: %v", err)
	}
	osFs := afero.NewOsFs()
	err = afero.WriteFile(osFs, "/tmp/nonexistingdir/testfile.txt", []byte("content"), 0777)
	if err != nil {
		log.Printf("write OsFs file: %v", err)
	}
}
```

Output from go run:

```write OsFs file: open /tmp/nonexistingdir/testfile.txt: no such file or directory```

### PR DESCRIPTION
**Pull request by veeceey**
Fixes #270

`MemMapFs.OpenFile` with `O_CREATE` was silently creating parent directories that don't exist. This meant `WriteFile` would happily write to non-existent paths in MemMapFs while `OsFs` would correctly return an error -- inconsistent behavior.

The fix adds a parent directory existence check in `OpenFile` before calling `Create`. If the parent doesn't exist, it returns `os.ErrNotExist`, matching `os.OpenFile` semantics.

Updated existing tests that relied on the implicit directory creation to explicitly create directories first (which is what you'd need to do on a real filesystem anyway). Added a dedicated test covering the bug.

### DIFF
--- a/afero_test.go
+++ b/afero_test.go
@@ -45,6 +45,8 @@ func testDir(fs Fs) string {
 }
 
 func tmpFile(fs Fs) File {
+	// Ensure the temp directory hierarchy exists in the filesystem.
+	fs.MkdirAll(os.TempDir(), 0o700)
 	x, err := TempFile(fs, "", "afero")
 	if err != nil {
 		panic(fmt.Sprint("unable to work with temp file", err))
--- a/ioutil_test.go
+++ b/ioutil_test.go
@@ -16,6 +16,7 @@
 package afero
 
 import (
+	"os"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -33,6 +34,7 @@ func checkSizePath(t *testing.T, path string, size int64) {
 
 func TestReadFile(t *testing.T) {
 	testFS = &MemMapFs{}
+	testFS.MkdirAll(os.TempDir(), 0o700)
 	fsutil := &Afero{Fs: testFS}
 
 	testFS.Create("this_exists.go")
@@ -53,6 +55,7 @@ func TestReadFile(t *testing.T) {
 
 func TestWriteFile(t *testing.T) {
 	testFS = &MemMapFs{}
+	testFS.MkdirAll(os.TempDir(), 0o700)
 	fsutil := &Afero{Fs: testFS}
 	f, err := fsutil.TempFile("", "ioutil-test")
 	if err != nil {
@@ -83,6 +86,7 @@ func TestWriteFile(t *testing.T) {
 
 func TestReadDir(t *testing.T) {
 	testFS = &MemMapFs{}
+	testFS.MkdirAll(os.TempDir(), 0o700)
 	testFS.Mkdir("/i-am-a-dir", 0o777)
 	testFS.Create("/this_exists.go")
 	dirname := "rumpelstilzchen"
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by veeceey** [comment]
### Test Results

Full test suite passes:

```
$ go test -count=1 ./...
ok  	github.com/spf13/afero	4.528s
ok  	github.com/spf13/afero/mem	0.195s
ok  	github.com/spf13/afero/tarfs	0.648s
ok  	github.com/spf13/afero/zipfs	0.506s
```

Verified the fix with the new test `TestOpenFileNonExistentDirectory`:

```
$ go test -count=1 -v -run TestOpenFileNonExistentDirectory ./...
=== RUN   TestOpenFileNonExistentDirectory
--- PASS: TestOpenFileNonExistentDirectory (0.00s)
```

Also confirmed manually that `WriteFile` to a non-existent directory now returns `os.ErrNotExist`, matching `OsFs` behavior.

spf13:
The fix is correct — MemMapFs should have matched `os.OpenFile` semantics here from the start, and the implementation is clean.

However, this is a breaking change for downstream consumers and we cannot merge it without coordination. I checked the major consumers:

**Hugo (`gohugoio/hugo`)** breaks in at least 4 test files: `hugofs/rootmapping_fs_test.go` (multiple test functions), `hugolib/filesystems/basefs_test.go` (`TestStaticFs`, `TestStaticFsMultihost`, `setConfigAndWriteSomeFilesTo` helper), and `hugolib/integrationtest_builder.go` (`writeToFs`/`AddFiles`). The pattern is `afero.WriteFile` on multi-level paths with no preceding `MkdirAll`.

**Grafana k6** (`grafana/k6:lib/fsext/trimpathseparator_test.go`) explicitly tests that `WriteFile` succeeds without pre-created directories — directly the behavior this PR removes.

**Viper and Cobra** are safe (Viper uses `fs.Create()` not `WriteFile`; Cobra does not use MemMapFs).

Holding this until we can do a coordinated release that fixes Hugo's test files simultaneously. The migration path for callers is straightforward: add `MkdirAll(filepath.Dir(path), perm)` before `WriteFile`, or use `afero.WriteReader` which already does this correctly.

spf13:
The fix here is correct — MemMapFs should require parent directories to exist before creating a file, matching os.OpenFile semantics. However, this is an observable behavior change for existing users: code that currently does WriteFile(memFs, "/some/path/file.txt", ...) without a preceding MkdirAll will silently break. MemMapFs is widely used as a frictionless test double, and many codebases rely on this permissiveness.

Before merging we need to assess the blast radius. A few questions:
1. How common is this pattern in the wild? (A quick grep of open-source afero users would tell us.)
2. Should this be gated behind a version bump with a clear migration note?
3. Is there an opt-in flag (e.g. StrictMode) that would let users choose the behavior, rather than a hard cut?

The implementation and tests are good. This is a policy/impact question, not a correctness question.

PR #573, opened 2026-02-23
fix: add permission checks to MemMapFs Open/OpenFile

### LINKED ISSUE #150: File permissions don't get checked by Open
It's possible to use the memmap fs to create a file, chmod it to 0, and then open it without a problem.  When using memmap to simulate a filesystem in unit tests, this behaviour is surprising.

Adding this to memmap.go's MemMapFs.open() just before the final return does the trick

	if mem.GetFileInfo(f).Mode().Perm()&0444 == 0 {
		return nil, &os.PathError{Op: "open", Path: name, Err: os.ErrPermission}
	}

but we also need a m.Chmod(0666) in Create, which then matches the documentation for os.Create() : https://golang.org/pkg/os/#Create

After applying only these changes, the afero_test fails because of some more permission problems in tempfile and tempdir creation I haven't tracked down yet.

veeceey:
Fixes #150

Right now MemMapFs ignores file permissions completely -- you can chmod a file to 0 and still open/read it without any error. This makes it pretty much impossible to use MemMapFs for testing code that cares about file permissions.

This PR adds basic permission checking:

- `Open()` checks read permission bits (0444) before returning a handle
- `openWrite()` checks write permission bits (0222) before returning a handle
- `Create()` now defaults to 0666 permissions, matching what `os.Create()` does
- `Stat()` bypasses the permission check (uses internal `open()`) since real `stat(2)` doesn't require read permission either
- Auto-created parent directories in `registerWithParent` now get 0755 instead of 0, so they stay accessible

The permission model is simplified -- it checks any read/write bit being set rather than doing full uid/gid owner/group/other matching. This keeps things straightforward while still catching the common case of chmod 0 or testing read-only vs write-only scenarios.

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -57,6 +57,7 @@ func (m *MemMapFs) Create(name string) (File, error) {
 	name = normalizePath(name)
 	m.mu.Lock()
 	file := mem.CreateFile(name)
+	mem.SetMode(file, 0o666)
 	m.getData()[name] = file
 	m.registerWithParent(file, 0)
 	m.mu.Unlock()
@@ -114,7 +115,11 @@ func (m *MemMapFs) registerWithParent(f *mem.FileData, perm os.FileMode) {
 	parent := m.findParent(f)
 	if parent == nil {
 		pdir := filepath.Dir(filepath.Clean(f.Name()))
-		err := m.lockfreeMkdir(pdir, perm)
+		mkdirPerm := perm
+		if mkdirPerm == 0 {
+			mkdirPerm = 0o755
+		}
+		err := m.lockfreeMkdir(pdir, mkdirPerm)
 		if err != nil {
 			// log.Println("Mkdir error:", err)
 			return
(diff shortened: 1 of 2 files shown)


veeceey:
### Test Results

Full test suite passes:

```
$ go test -count=1 ./...
ok  	github.com/spf13/afero	4.358s
ok  	github.com/spf13/afero/mem	0.156s
ok  	github.com/spf13/afero/tarfs	0.454s
ok  	github.com/spf13/afero/zipfs	0.603s
```

New test `TestMemMapFsPermissionChecks` covers:

```
$ go test -v -run TestMemMapFsPermissionChecks ./...
=== RUN   TestMemMapFsPermissionChecks
--- PASS: TestMemMapFsPermissionChecks (0.00s)
PASS
```

Verified manually:
- Creating a file gives 0666 permissions (matching os.Create docs)
- Chmod to 0 then Open -> permission denied
- Chmod to 0444 then Open for read -> ok, Open for write -> denied
- Chmod to 0222 then Open for read -> denied
- Stat works regardless of permissions (matching real stat behavior)

spf13:
Applied to master. This is a meaningful step toward making MemMapFs behave correctly under permission-sensitive code.

PR #588, opened 2026-04-14
fix: return os.ErrClosed on double Close() in MemMapFs

### LINKED ISSUE #567: MemMapFS Does not Error on Double Close
The OsFS errors on double close whereas the MemMapFS does not. This leads to inconsistent behaviors between tests that use MemMapFS and production code that uses OsFS

```golang
package main

import (
	"testing"

	"github.com/spf13/afero"
	"github.com/stretchr/testify/assert"
)

func TestAferoMemMapFsDoubleClose(t *testing.T) {
	fs := afero.NewMemMapFs()
	f, err := fs.Create("test")
	assert.NoError(t, err)

	err = f.Close()
	assert.NoError(t, err)

	err = f.Close()
	assert.Error(t, err, "Should error on second close")
}

func TestAferoOsFSDoubleClose(t *testing.T) {
	fs := afero.NewBasePathFs(afero.NewOsFs(), t.TempDir())
	f, err := fs.Create("test")
	assert.NoError(t, err)

	err = f.Close()
	assert.NoError(t, err)

	err = f.Close()
	assert.Error(t, err, "Should error on second close")
}

```

Yanhu007:
Fixes #567

## Problem

`mem.File.Close()` silently succeeds when called multiple times, while `os.File.Close()` returns an error (`os.ErrClosed`). This inconsistency causes bugs to go undetected in tests using MemMapFs that only surface in production with OsFs.

## Fix

Check the `closed` flag under the lock before proceeding. If already closed, return `os.ErrClosed` to match `os.File` behavior.

## Testing

All existing tests pass.

### DIFF
--- a/mem/file.go
+++ b/mem/file.go
@@ -128,6 +128,10 @@ func (f *File) Open() error {
 
 func (f *File) Close() error {
 	f.fileData.Lock()
+	if f.closed {
+		f.fileData.Unlock()
+		return os.ErrClosed
+	}
 	f.closed = true
 	if !f.readOnly {
 		setModTime(f.fileData, time.Now())


spf13:
Closing as duplicate of #590, which fixes the same issue (#567), has a signed CLA, and uses defer for cleaner lock management. @Yanhu007 please sign the CLA (https://cla-assistant.io/spf13/afero) so we can move #587 forward — that one is also yours and it's a good fix.

PR #589, opened 2026-04-15
opt: Supports returning all file names

wuxia101:


spf13:
Closing. There are two issues:

1. **Missing mutex.** `ListNames()` iterates `m.getData()` without holding `m.mu`. This is a data race against any concurrent write operation. The fix requires wrapping the loop in `m.mu.RLock() / m.mu.RUnlock()`.

2. **API design.** Adding `ListNames()` alongside the existing `List()` (which prints to stdout — a design mistake from day one) makes the API messier. The right fix is to clean up `List()` itself, which we'll do in a separate pass. In the meantime, `Walk()` is the correct tool for enumerating all files programmatically.

wuxia101:
Thank you for taking the time to review this issue , Thank you.  When I use an LLM to operate Memefs, only the file name is required, and my agent will handle the file operations automatically. in case，the best way。

PR #597, opened 2026-05-09
mem: return error when reading a directory (fixes #430)

### LINKED ISSUE #430: memfs directory can be read by ReadFile
When you have a test doing something like this:

```go
	configFile := filepath.Join(t.TempDir(), "config")
	_ = os.Mkdir(configFile, 0755)
```

and the file under tests does a `afero.ReadFile(fs, configPath)`, the call returns without an error and an empty result instead of failing on the fact that the configPath is a directory.

The same code works as expected on `osfs`

nghiack7:
## Summary

`afero.ReadFile(fs, dirPath)` on a `MemMapFs` silently returned empty content with a `nil` error. The real OS filesystem returns an error (`read /some/dir: is a directory`). This inconsistency was reported in #430.

### Root cause

`(*File).Read` delegates to `ReadAt`, which had no guard for directory-backed files. It would simply return `n=0, err=nil` because the file data slice is empty for directories.

### Fix

Added a directory check at the top of `ReadAt`:

```go
if f.fileData.dir {
    return 0, &os.PathError{Op: "read", Path: f.fileData.name, Err: errors.New("is a directory")}
}
```

`Read` already delegates to `ReadAt`, so both are covered by one guard.

### Test

Added `TestReadFileOnDirectory` in `util_test.go` that:
1. Creates a `MemMapFs` directory
2. Calls `afero.ReadFile` on it
3. Asserts a non-nil error is returned

```
$ go test ./... -run TestReadFile
ok      github.com/spf13/afero  0.6s
```

Full test suite passes without regressions.

Fixes #430

### DIFF
--- a/mem/file.go
+++ b/mem/file.go
@@ -219,6 +219,9 @@ func (f *File) ReadAt(b []byte, off int64) (n int, err error) {
 	if f.closed {
 		return 0, ErrFileClosed
 	}
+	if f.fileData.dir {
+		return 0, &os.PathError{Op: "read", Path: f.fileData.name, Err: errors.New("is a directory")}
+	}
 	if len(b) > 0 && int(off) == len(f.fileData.data) {
 		return 0, io.EOF
 	}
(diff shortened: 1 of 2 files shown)


nghiack7:
@CLAassistant recheck

PR #601, opened 2026-05-17
fix: propagate errors from CacheOnReadFs.Open

### LINKED ISSUE #116: CacheOnReadFs.Open() is dropping errors
At the bottom of the function body are these lines:

```go
	// the dirs from cacheHit, cacheStale fall down here:
	bfile, _ := u.base.Open(name)
	lfile, err := u.layer.Open(name)
	if err != nil && bfile == nil {
		return nil, err
	}
	return &UnionFile{base: bfile, layer: lfile}, nil
```

The second line in the above snippet drops the error (from `u.base.Open(name)`), which means that any implementation which returns a valid interface and an error (for example, an interface object implemented by a nil pointer whose type happens to satisfy the File interface) will result in a probably-corrupt UnionFile (in the case of a non-nil interface object implemented by a nil pointer, calling `UnionFile.Readdir()` will panic because the method only checks that the interface object is non-nil).

I think this could be fixed via:

```go
bfile, berr := u.base.Open(name)
lfile, lerr := u.layer.Open(name)
if lerr != nil && berr != nil { // check the error, not the value
    return nil, lerr
}
return &UnionFile{base: bfile, layer: lfile}, nil
```

Although it's probably bad practice to silently swallow non-FileNotFound base errors--if my base FS system is under duress, I probably don't want to pretend the cache is the new source of truth.

wyf027:
## Summary

Fixes #116.

`CacheOnReadFs.Open` no longer ignores errors from `base.Open` when opening unioned directories. Both base and layer errors are returned to callers instead of constructing a potentially invalid `UnionFile`.

## Test plan

- [x] `go test ./...`

### DIFF
--- a/cacheOnReadFs.go
+++ b/cacheOnReadFs.go
@@ -271,9 +271,12 @@ func (u *CacheOnReadFs) Open(name string) (File, error) {
 		}
 	}
 	// the dirs from cacheHit, cacheStale fall down here:
-	bfile, _ := u.base.Open(name)
+	bfile, err := u.base.Open(name)
+	if err != nil {
+		return nil, err
+	}
 	lfile, err := u.layer.Open(name)
-	if err != nil && bfile == nil {
+	if err != nil {
 		return nil, err
 	}
 	return &UnionFile{Base: bfile, Layer: lfile}, nil


spf13:

PR #604, opened 2026-05-17
fix(mem): implement io.ReaderFrom on MemMapFs files

### LINKED ISSUE #498: Support for afero.File.ReadFrom (to implement io.ReaderFrom)
`os.file` has the following function:

```
// ReadFrom implements io.ReaderFrom.
func (f *File) ReadFrom(r io.Reader) (n int64, err error) {
	if err := f.checkValid("write"); err != nil {
		return 0, err
	}
	n, handled, e := f.readFrom(r)
	if !handled {
		return genericReadFrom(f, r) // without wrapping
	}
	return n, f.wrapErr("write", e)
}
```

Is it something we can have in `afero`?

wyf027:
## Summary

- Implements `io.ReaderFrom` on `mem.File` so `MemMapFs` files behave like `*os.File` for `io.Copy` fast paths.
- `ReadFrom` writes starting at the current file offset and advances it, matching `os.File` semantics.
- Adds unit tests in `mem` and integration tests for `MemMapFs`.

Fixes #498

## Test plan

- [x] `go test ./...`

### DIFF
--- a/mem/file.go
+++ b/mem/file.go
@@ -30,7 +30,10 @@ import (
 
 const FilePathSeparator = string(filepath.Separator)
 
-var _ fs.ReadDirFile = &File{}
+var (
+	_ fs.ReadDirFile = &File{}
+	_ io.ReaderFrom  = (*File)(nil)
+)
 
 type File struct {
 	// atomic requires 64-bit alignment for struct field access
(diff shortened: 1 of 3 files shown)


spf13:

PR #607, opened 2026-05-25
fix(mem): align MemMapFs Stat with os.FileInfo behavior

### LINKED ISSUE #222: HttpFs.Dir doesn't work properly when given root path
If you try
```
httpFs := afero.NewHttpFs(<ExistingFS>)
fileserver := http.FileServer(httpFs.Dir("/")))
http.Handle("/", fileserver)
```
you'll find that your fileserver successfully lists the directory, but gives 404 on any file. However, if you do
```
httpFs := afero.NewHttpFs(<ExistingFS>)
fileserver := http.FileServer(httpFs.Dir("/something/")))
http.Handle("/", fileserver)
```
everything works as expected.

### LINKED ISSUE #522: MemMapFs.Stat does not handle empty string as expected
os.Stat("") will return an error, `stat : no such file or directory`, where as MemMapFs.Stat() does not return an error. Instead it returns a populated FileInfo:
```
FileInfo.Name() == ""
FileInfo.Size() == 42
FileInfo.Mode() == "drwxr-xr-x"
FileInfo.ModTime() == "2025-07-30 16:15:24.211543491 -0400 EDT m=+0.000353724"
FileInfo.IsDir() == true
```

This can be worked around by having special case handling for empty strings in the code calling into Stat(), but that kinda defeats the purpose of having an injectable FS for testing.

A test showing this difference in behaviour:
```go
func TestMemMapFsStat(t *testing.T) {
       fs := &MemMapFs{}
       {
               f, err := fs.Stat("")
               if err != nil {
                       fmt.Printf("fs.Stat() err is %s\n", err.Error())
               } else {
                       fmt.Println("fs.Stat() err is nil")
                        fmt.Println(f.Name())
                        fmt.Println(f.Size())
                        fmt.Println(f.Mode())
                        fmt.Println(f.ModTime())
                        fmt.Println(f.IsDir())
               }
       }
       {
               f, err := os.Stat("")
               if err != nil {
                       fmt.Printf("os.Stat() err is %s\n", err.Error())
               } else {
                       fmt.Println("os.Stat() err is nil")
                        fmt.Println(f.Name())
                        fmt.Println(f.Size())
                        fmt.Println(f.Mode())
                        fmt.Println(f.ModTime())
                        fmt.Println(f.IsDir())
               }
       }
}
```

wyf027:
## Summary

- `MemMapFs.Stat("")` now returns a `*os.PathError` with `Op: "stat"` instead of surfacing an internal `open` error
- `FileInfo.Name()` uses `filepath.Base`, so `Stat("/")` reports `"/"` like `os.Stat` instead of `""`

## Problem

Injectable filesystem tests comparing against `os.Stat` were tripped up by two MemMapFs divergences:

1. Empty path stat errors used `Op: "open"` (issue #522)
2. Root directory `Name()` was `""` because `filepath.Split("/")` yields an empty base name (issue #222)

## Test plan

- [x] Added `TestMemMapFsStatEmptyPath`
- [x] Added `TestMemMapFsStatRootName`
- [x] `go test ./...`

Fixes #522
Fixes #222

### DIFF
--- a/mem/file.go
+++ b/mem/file.go
@@ -340,9 +340,9 @@ type FileInfo struct {
 // Implements os.FileInfo
 func (s *FileInfo) Name() string {
 	s.Lock()
-	_, name := filepath.Split(s.name)
+	name := s.name
 	s.Unlock()
-	return name
+	return filepath.Base(name)
 }
 
 func (s *FileInfo) Mode() os.FileMode {
--- a/memmap.go
+++ b/memmap.go
@@ -402,8 +402,14 @@ func (m *MemMapFs) LstatIfPossible(name string) (os.FileInfo, bool, error) {
 }
 
 func (m *MemMapFs) Stat(name string) (os.FileInfo, error) {
+	if name == "" {
+		return nil, &os.PathError{Op: "stat", Path: name, Err: ErrFileNotFound}
+	}
 	f, err := m.open(name)
 	if err != nil {
+		if pe, ok := err.(*os.PathError); ok {
+			return nil, &os.PathError{Op: "stat", Path: pe.Path, Err: pe.Err}
+		}
 		return nil, err
 	}
 	fi := mem.GetFileInfo(f)
--- a/memmap_stat_test.go
+++ b/memmap_stat_test.go
@@ -1,0 +1,41 @@
+package afero
+
+import (
+	"errors"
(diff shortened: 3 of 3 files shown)


spf13:

PR #610, opened 2026-05-25
fix(mem): return EISDIR when reading from directory handles

### LINKED ISSUE #201: afero.ReadFile doesn't return error when filename is a directory when fs is memmapfs

wyf027:
## Summary

Fixes #201

`afero.ReadFile` on a directory path with `MemMapFs` previously returned empty data without error. The in-memory file handle allowed `Read`/`ReadAt` on directory nodes.

This change aligns `mem.File` with `tarfs`, `zipfs`, and `gcsfs` by returning a `PathError` wrapping `syscall.EISDIR` for read, write, seek, and truncate operations on directories.

## Test plan

- [x] `go test ./...`
- [x] Added `TestReadFileDirectoryMemMapFs`

### DIFF
--- a/ioutil_test.go
+++ b/ioutil_test.go
@@ -16,8 +16,11 @@
 package afero
 
 import (
+	"errors"
+	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


spf13:

PR #616, opened 2026-05-25
fix: use case-insensitive path lookup in MemMapFs on Windows

### LINKED ISSUE #126: Stat call when using Memory FS is case sensitive on Windows
Don't you love Windows?  I'm not even sure if this is an issue, your call.  I'm trying to make my library Windows happy so I'm unit testing with Windows paths.  When I create an in-memory filesystem and look for a file, the call Stat relies eventually on the internal call "open" which does a map lookup on the result of getData, using the filename as a key.  While on many operating systems this behavior is completely reasonable, on some (cough, cough, Windows) it would need to be a case independent compare.  Is this something that qualifies as an issue or am I misusing and abusing the in memory filesystem for unholy purposes?

wyf027:
## Summary

- MemMapFs used exact-case map keys for path lookup, which breaks Windows-style paths where casing may differ between create and stat/open calls
- Add a shared lookup helper that falls back to `strings.EqualFold` on Windows while preserving case-sensitive behavior elsewhere

## Test plan

- [x] Added `TestMemMapFsCaseInsensitiveLookup` (Windows vs non-Windows behavior)
- [x] `go test ./...`

Fixes #126

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -19,6 +19,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


spf13:

PR #618, opened 2026-05-25
fix: treat nil MemMapFs FileData entries as missing

### LINKED ISSUE #421: [MemMapFs] Check for getting nil `mem.FileData` by key
About `v.1.11.0`.

Static analysis shows two critical:
1. > After having been assigned to a nil value at memmap.go:240, pointer '&parent.Mutex' is passed as implicit 'this' parameter in call to function 'sync.Mutex.Lock' at memmap.go:131, where it is dereferenced at mutex.go:16.
2. > After having been assigned to a nil value at memmap.go:240, pointer 'f' is passed as 1st parameter in call to function 'afero.MemMapFs.findParent' at memmap.go:73, where it is dereferenced at memmap.go:85.

It's looks like false positive because an error returns on line 240 that is being handled in methods `registerWithParent` and `unRegisterWithParent`. But there is a moment. `getData()` returns a map that stores pointers and there is no check for getting `nil` by existed keys in methods `lockfreeOpen` and `open`. This potentially may cause panics.

Please add checks for getting `nil` where they need to be.

wyf027:
## Summary

`MemMapFs` stores files in a `map[string]*mem.FileData`. A map lookup can return `ok == true` with a nil pointer value, which previously caused panics when dereferencing the result in `open`, `lockfreeOpen`, and metadata helpers.

This change treats nil entries the same as missing keys and returns not-found / `ErrFileNotFound` errors.

Fixes #421

## Test plan

- [x] `go test ./...`
- [x] Added `TestMemMapFsNilFileDataEntry` covering `Open` and `Stat` with a manually inserted nil map entry

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -140,7 +140,7 @@ func (m *MemMapFs) registerWithParent(f *mem.FileData, perm os.FileMode) {
 func (m *MemMapFs) lockfreeMkdir(name string, perm os.FileMode) error {
 	name = normalizePath(name)
 	x, ok := m.getData()[name]
-	if ok {
+	if ok && x != nil {
 		// Only return ErrFileExists if it's a file, not a directory.
 		i := mem.FileInfo{FileData: x}
 		if !i.IsDir() {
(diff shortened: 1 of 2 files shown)


spf13:

PR #619, opened 2026-05-25
refactor: remove unused base parameter from copyFile

### LINKED ISSUE #422: Unused parameter in copyFile function in unionFile
In unionFile.go, at line no 270, copyFile function accepts a parameter named "base" https://github.com/spf13/afero/blob/5c4385aa20510dba5ca4db12c02b0c9211d82892/unionFile.go#L270C15-L270C19, which is not used.

Could be a good practice to remove it from the function definition. I could raise a PR for the same once issue is accepted.

wyf027:
## Summary

`copyFile` accepted a `base Fs` parameter that was never used inside the function. Callers already open the source file from `base` before delegating to `copyFile`.

Fixes #422

## Test plan

- [x] `go test ./...`

### DIFF
--- a/unionFile.go
+++ b/unionFile.go
@@ -274,7 +274,7 @@ func (f *UnionFile) WriteString(s string) (n int, err error) {
 	return 0, BADFD
 }
 
-func copyFile(base Fs, layer Fs, name string, bfh File) error {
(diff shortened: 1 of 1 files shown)


spf13:

PR #620, opened 2026-05-25
fix: only enforce O_EXCL on MemMapFs when O_CREATE is set

### LINKED ISSUE #415: Bug: OpenFile broken for MemMapFs
Basic scenario doesn't work.

1. Create a file with os.O_CREATE|os.O_RDWR|os.O_EXCL and close it
2. Try to reopen the file with os.O_RDWR|os.O_EXCL

Result: file exists error

Expected: reopen should succeed

See

https://goplay.tools/snippet/e6a3inm51RX

wyf027:
## Summary

`MemMapFs.OpenFile` treated `O_EXCL` as "file must not exist" on every open. That breaks reopens such as `O_RDWR|O_EXCL` after the file was already created, while the real OS ignores `O_EXCL` unless `O_CREATE` is also set.

Only return `ErrFileExists` when both `O_CREATE` and `O_EXCL` are present and the file already exists.

Fixes #415

## Test plan

- [x] `go test ./...`
- [x] Existing `TestOpenFileExcl` still passes (create+excl twice fails as expected)
- [x] Added `TestOpenFileExclWithoutCreate` for create then reopen with `O_RDWR|O_EXCL`

### DIFF
--- a/memmap.go
+++ b/memmap.go
@@ -263,7 +263,7 @@ func (m *MemMapFs) OpenFile(name string, flag int, perm os.FileMode) (File, erro
 	perm &= chmodBits
 	chmod := false
 	file, err := m.openWrite(name)
-	if err == nil && (flag&os.O_EXCL > 0) {
+	if err == nil && (flag&os.O_CREATE > 0) && (flag&os.O_EXCL > 0) {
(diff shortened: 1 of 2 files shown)


spf13:

PR #637, opened 2026-07-05
fix: forward syscall.Conn through BasePathFile and RegexpFile

### LINKED ISSUE #630: Wrapper files (`BasePathFile`, etc) hide `*os.File`'s `syscall.Conn`, defeating `net/http`'s `sendfile(2)` fast path
`net/http.ServeContent` or `io.Copy` to a `net.TCPConn` only uses the zero-copy `sendfile(2)` path when the source reader implements `syscall.Conn`. See [net/sendfile.go](https://cs.opensource.google/go/go/+/refs/tags/go1.26.4:src/net/sendfile.go;l=38-41). `*os.File` implements `syscall.Conn`, so serving a file straight from `os.Open` gets the `sendfile(2)` optimization. If you're not aware, `sendfile(2)` [has a *big* impact on serving performance](https://thelinuxcode.com/nginx-sendfile-directive/).

But when a file passes through a wrapping Fs (`BasePathFs`, `ReadOnlyFs`, `RegexpFs`, etc.) the wrapper file type does not expose `syscall.Conn` so `net.sendFile` falls back to a userspace read/write copy, losing the optimization. `BasePathFs` is a very common base, so in practice afero silently disables `sendfile(2)` for most static-file serving cases.

---

As a workaround I found that recursively type asserting `afero.File` into an `afero.BasePathFile` to access its underlying `File` field until I get an `*os.File` works ok in my case (see `func osFile` below), but this is awkward, undocumented, and not universal.

```go
func osFile(f afero.File) (*os.File, bool) {
	for {
		switch v := f.(type) {
		case *os.File:            return v, true
		case *afero.BasePathFile: f = v.File
		default:                  return nil, false
		}
	}
}
```

---

Ideally users that use afero and don't know about sendfile(2) wouldn't have to pay a performance penalty when serving files over the network. Perhaps afero could return a different File type that does implement `syscall.Conn`. But then, whether the wrapper type implements `syscall.Conn` should depend on whether the wrapped type does, which in turn depends on exactly which file is opened because you could land on an afero layer that doesn't wrap an os filesystem at all like memmapfs which can't support sendfile/`syscall.Conn`. I don't think it would be too hard to do this just for the `syscall.Conn` interface, just recursively unwrap the file and if the bottom file type implements `syscall.Conn` then return an alternative afero type that also implements the interface in that case.

Note that to fully mirror the wrapped type you'd have to forward every optional interface it implements. For files that's a handful (`syscall.Conn`, `io.ReaderFrom`, `io.WriterTo`, …) but not nothing. It's the same problem [httpsnoop.Wrap](https://pkg.go.dev/github.com/felixge/httpsnoop#Wrap) solves for `http.ResponseWriter` by generating 2^n implementations and picking the one that perfectly mirrors the source. Not asking for that here, `syscall.Conn` alone fixes my case.

A simpler but less smooth alternative is just explicitly documenting and canonicalizing the workaround above in a free function, say `afero.UnwrapOsFile(afero.File) (*os.File, bool)` or something more specific like `afero.UnwrapSyscallConnFile(afero.File) (afero.File, bool)` (where the returned afero.File may implement `syscall.Conn` if the underlying file does, forwarding calls to the unwrapped File type).

---

My preference would be `BasePathFs.Open()` etc returns a type that implements `syscall.Conn` *if* the backing fs does. Barring that, I'd be satisfied with a func that canonicalizes the workaround.

---

References:

- https://github.com/golang/go/issues/69097
- https://github.com/golang/go/issues/9113
- https://segflow.github.io/post/zero-copy-sendfile-splice/

momomuchu:
## Summary
- add SyscallConn on BasePathFile, forwarding to the wrapped file when it implements syscall.Conn
- add SyscallConn on RegexpFile, same pattern
- regression tests for both, plus negative-control tests for wrapped files that don't support it (MemMapFs)

## Why
*os.File implements syscall.Conn, which net/http and io.Copy use to hit the sendfile(2) fast path. BasePathFile and RegexpFile don't forward it, so anything served through BasePathFs or RegexpFs silently falls back to a userspace copy. Same class of bug as #577 (WriterTo/ReaderFrom), just a different interface.

ReadOnlyFs.Open returns the source file directly without wrapping, so it already passes syscall.Conn through and needs no change here.

CopyOnWriteFs wraps into UnionFile{Base, Layer}, not a single-file pass-through, so forwarding syscall.Conn there is a separate design question (which side wins) and out of scope for this PR.

Not overlapping with #632 (WriterTo/ReaderFrom on RegexpFile), different interface on the same wrapper, filed independently.

## Test Plan
- go test ./...
- go test -race ./...

Fixes #630

### DIFF
--- a/basepath.go
+++ b/basepath.go
@@ -1,12 +1,14 @@
 package afero
 
 import (
+	"errors"
 	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
+	"syscall"
 	"time"
 )
 
@@ -15,6 +17,7 @@ var (
 	_ fs.ReadDirFile = (*BasePathFile)(nil)
 	_ io.WriterTo    = (*BasePathFile)(nil)
 	_ io.ReaderFrom  = (*BasePathFile)(nil)
+	_ syscall.Conn   = (*BasePathFile)(nil)
 )
 
 // The BasePathFs restricts all operations to a given path within an Fs.
@@ -59,6 +62,13 @@ func (f *BasePathFile) ReadFrom(r io.Reader) (int64, error) {
 		return rf.ReadFrom(r)
 	}
 	return io.Copy(f.File, r)
+}
+
+func (f *BasePathFile) SyscallConn() (syscall.RawConn, error) {
+	if sc, ok := f.File.(syscall.Conn); ok {
+		return sc.SyscallConn()
+	}
+	return nil, errors.New("afero: underlying file does not implement syscall.Conn")
 }
 
 func NewBasePathFs(source Fs, path string) Fs {
--- a/regexpfs.go
+++ b/regexpfs.go
@@ -1,12 +1,15 @@
 package afero
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
 	"regexp"
 	"syscall"
 	"time"
 )
+
+var _ syscall.Conn = (*RegexpFile)(nil)
 
 // The RegexpFs filters files (not directories) by regular expression. Only
 // files matching the given regexp will be allowed, all others get a ENOENT error (
@@ -221,4 +224,11 @@ func (f *RegexpFile) Truncate(s int64) error {
 
 func (f *RegexpFile) WriteString(s string) (int, error) {
 	return f.f.WriteString(s)
+}
+
+func (f *RegexpFile) SyscallConn() (syscall.RawConn, error) {
+	if sc, ok := f.f.(syscall.Conn); ok {
+		return sc.SyscallConn()
+	}
+	return nil, errors.New("afero: underlying file does not implement syscall.Conn")
 }
--- a/syscall_conn_test.go
+++ b/syscall_conn_test.go
@@ -1,0 +1,122 @@
+package afero
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"syscall"
+	"testing"
+)
+
+// controlSucceeds runs rc.Control and fails the test if either the Control
+// call itself or the callback-reported syscall error is non-nil, and if the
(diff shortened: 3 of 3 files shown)


infogulch:
Oh cool I didn't think about just always implementing the interface and returning an error if the underlying file type doesn't implement `syscall.Conn`. That's a great simplifying approach which is obvious in hindsight. :)

---

The `CopyOnWriteFs` case which returns `UnionFile{Base, Layer}` is a bit tricky as you note. If the caller gets a [`RawConn`](https://pkg.go.dev/syscall#RawConn) then they can use any of `Read`, `Write`, and `Control`. So if you default to `Base` and the caller uses `Write`, that violates the expectations of `CopyOnWriteFs`. And if you default to `Layer` but the file only exists in `Base` then calling `Read` would fail unexpectedly.

Maybe the right answer is to kick the can down the road and return a new `UnionRawConn{Base, Layer}` type which forwards calls to the `RawConn` depending on the method+existence state:

* Exists in `Layer`: always resolve to `Layer`
* Doesn't exist in `Layer`, does exist in `Base` (maybe double check how layered deletions work)
   * Method `Read`: Forward to `Base`
   * Method `Write` or `Control`: Copy from `Base` to `Layer` then forward to `Layer`

I guess there would still be two issues to resolve:

* Inconsistent `syscall.Conn` support between the layers. If you layer a memoryfs on top of a basepathfs, then you could get a `UnionRawConn` and successfully call `Read`, but it will always fail when you try to call `Write` or `Control` because the memoryfs will never support RawConn. Maybe this is fine.
* When a caller gets an fd passed to its func in `Read(f func(fd uintptr) bool) error)`, how do you enforce that the caller only does read-only actions on the file? Ensure to open with `O_RDONLY`? Maybe that's enough.

PR #644, opened 2026-07-23
fix: nil BasePathFs doesn't panic on Stat

sonnemusk:
```go
var b *afero.BasePathFs
b.Stat("x") // panic
```

RealPath returns a PathError when the receiver is nil.


sonnemusk:
Closing this — it was a defensive nil guard, not something I hit in real use. Callers should not pass nil here; panicking (or failing at the call site) is fine.

PR #645, opened 2026-07-23
fix: reject empty path in ReadFile and WriteFile

sonnemusk:
## What

`ReadFile` / `WriteFile` reject empty filenames with `os.PathError` / `os.ErrInvalid`.

## Why

Matches `os` package behavior and avoids surprising operations on an empty path.

## Test

- `TestWriteReadFileEmptyPath`

### DIFF
--- a/empty_path_test.go
+++ b/empty_path_test.go
@@ -1,0 +1,23 @@
+package afero
+
+import (
+	"errors"
+	"os"
+	"testing"
+)
+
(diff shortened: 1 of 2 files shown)


sonnemusk:
Closing this PR. After review, this is too low-value / looks like bulk automated noise rather than a carefully chosen change. Sorry for the noise — will focus on real, high-signal fixes only.

